PR #9, opened 2020-12-01
copy_to_glacier Unit Tests

AndrewDorn:
## Summary of Changes

Added unit tests to copy_to_glacier and updated documentation.

Addresses [Create/update tests for copy_to_glacier lambda to ensure coverage](https://bugs.earthdata.nasa.gov/browse/ORCA-63)

## Changes

* Added unit tests around handler.py
* Updated docs

## Type of change

## How Has This Been Tested?

Unit tests pass on my machine.

## Checklist:

    - [x] Testing documentation
    - [x] Snyk
    - [x] git-secrets

### DIFF
--- a/tasks/copy_to_glacier/test/test_handler.py
+++ b/tasks/copy_to_glacier/test/test_handler.py
@@ -1,6 +1,9 @@
 import json
+import uuid
 from os import path
 from unittest import TestCase
+from unittest.mock import Mock, call
+
 from ..handler import *
 
 
@@ -9,6 +12,7 @@ class TestCopyToGlacierHandler(TestCase):
     Test copy to glacier handler
     This will test if the function of test copy to glacier work as expected
     """
+
     excluded_file = 's3://test-bucket/this_file_should_be_exclude.example'
     not_excluded_file = 's3://test-bucket/prefix/this_file_should_not_be_exclude.txt'
     two_prefix_file = 's3://test-bucket/prefix1/prefix2/file.txt'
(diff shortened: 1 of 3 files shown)


tasks/copy_to_glacier/test/test_handler.py
```diff
@@ -54,15 +58,420 @@ def test_4_get_bucket(self):
...
         """
-        Testing task
+        Basic path with buckets present.
         """
-        context = []
-        with open(self.event_path) as f:
-            event = json.load(f)
-            test_handler = task(event, context)
-            print(test_handler)
+        collection_name = uuid.uuid4().__str__()
```
ifestus, line 68: I think it would make sense to move this block of variable initialization into its own helper function. That should cut down on a decent amount of the duplicate code.

ifestus: Another option would be to just initialize one event object and pass a copy of that object to each test. If you don't have anything relying on the attributes being different for each test, that is. <- this can be done with the `setUp` and `tearDown` functions.

tasks/copy_to_glacier/test/test_handler.py
```diff
@@ -54,15 +58,420 @@ def test_4_get_bucket(self):
...
+            ExtraArgs={
+                'StorageClass': 'GLACIER',
+                'MetadataDirective': 'COPY',
+                'ContentType': content_type
+            }
+        )])
+
+    def test_task_bucket_not_present(self):
+        """
+        Bucket not present in collection, should default to 'public' in output.
```
ifestus, line 145: Good description. I was just about to ask about this before seeing the above line =).

ifestus (commented):
I have just a few comments/suggestions, but this looks really good so far!

tasks/copy_to_glacier/test/test_handler.py
ifestus: https://github.com/nasa/cumulus-orca/blob/develop/tasks/copy_files_to_archive/test/test_copy_files_to_archive.py#L28 setup and teardown example

AndrewDorn: Sadly, the event does have to be different. Both in terms of raw values and in terms of attributes being present or not present. I could have a helper function that gets passed a bunch of bools (set_property = false, set_another_property = false, set_a_third_property = false) but that might get a bit messy.
Also note that we need to know what these values are when writing the asserts, so we'd need to pass a bool into the helper function along with a value.

ifestus: I think I see where you're coming from. I was thinking about a helper function that just returns a dictionary of uuid4s so you'd be able to pick and choose which ones to use, but the trade-off is that you have to index into the dictionary for each variable reference. With the context you provided I think its fine to ignore my previous suggestions.

ifestus (approved):
LGTM 👍

PR #28, opened 2020-12-29
[ORCA-67] Generalize input/output of copy_to_glacier lambda so it can be used anywhere after MoveGranules in a workflow

ifestus:
## Summary of Changes

Generalize input/output of copy_to_glacier lambda so it can be used anywhere in a workflow.

Addresses [ORCA-67: Generalize input/output of copy_to_glacier lambda so it can be used anywhere in a workflow.](https://bugs.earthdata.nasa.gov/browse/ORCA-67)

## Changes

* `copy_to_glacier` function now takes a granule object, like the one returned from Cumulus' `MoveGranules` as input.
* Removed a bunch of code that was used for generating filepaths/meta information based on the s3 filepath. Those values are all part of the input `granule` object.
* Remove support for setting a custom staging directory - files in Glacier storage will have the same s3 key as those in primary storage - only the bucket is different.
* Update tests and `README.md`.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

- [x] Unit tests
- [x] Manual integration testing with the orca-sandbox stack

## Checklist:

    - [x] CHANGELOG.md
    - [x] Testing documentation
    - [x] Unit tests
    - [x] Snyk
    - [x] git-secrets

Currently the input to `copy_to_glacier` is a list of s3 filepaths. This is simple, but also means that users somehow have to generate this string in some previous Cumulus task.

My changes take the input of a standard Cumulus task as output so the `copy_to_glacier` task is easier to integrate into a workflow.

### DIFF
--- a/tasks/copy_to_glacier/handler.py
+++ b/tasks/copy_to_glacier/handler.py
@@ -84,7 +84,6 @@ def task(event: Dict[str, Union[List[str], Dict]], context: object) -> Dict[str,
     """
     print(event)
     event_input = event.get('input')
-    # If there is no granules object, fail the workflow.
     granules_list = event_input.get('granules')
     config = event.get('config')
     collection = config.get(CONFIG_COLLECTION_KEY)
--- a/tasks/copy_to_glacier/handler.py
+++ b/tasks/copy_to_glacier/handler.py
@@ -57,54 +57,6 @@ def copy_granule_between_buckets(source_bucket_name: str, source_key: str, desti
         }
     )
 
-
-# todo: Once aws upgrades to Python 3.8, have this return Optional[Match[AnyStr]]
-def get_source_bucket_and_key(granule_url) -> Optional[Match]:
-    """
-    Parses source bucket and key from s3 url.
-    Args:
-        granule_url: s3 url path to granule.
-    Returns:  TODO: Strip 0th element from return value.
-        re.Match object with argument [1] equal to source bucket name and [2] equal to source key.
-    """
-    return re.search("s3://([^/]*)/(.*)", granule_url)
-
-
-def get_bucket_name_for_filename(filename: str, collection_files: List[Dict[str, Any]]) -> str:
-    """
-    Retrieves the first file pattern in {collection_files} where the file's ['regex'] matches the {filename}
-    And returns that file's ['bucket']
-    Args:
-        filename: Granule file name.
-        collection_files: List of collection files.
-            Each file is a dict with the following keys:
-                regex (str): The regex that all files in the bucket must match with their name.
-                bucket (str): The name of the bucket containing the files.
-    Returns:
-        Bucket name, or 'public' if not found.
-    """
-    for file in collection_files:
-        if re.match(file.get('regex', '*.'), filename):
-            return file['bucket']
-    return 'public'
-
-
-def get_granule_urls_from_granules_list(granules_list: List) -> List[str]:
-    """
-    Pulls a list of filenames (S3 filepaths) from a list of granules.
-
-    Args:
-        granules_list: List of granule objects
-
-    Returns:
-        List of S3 filepaths.
-    """
-    files = []
-    for granule in granules_list:
-        files.extend([ file['filename'] for file in granule['files'] ])
-    return files
-
-
 # noinspection PyUnusedLocal
 def task(event: Dict[str, Union[List[str], Dict]], context: object) -> Dict[str, Any]:
     """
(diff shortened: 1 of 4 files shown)


tasks/copy_to_glacier/handler.py
```diff
@@ -115,42 +83,37 @@ def task(event: Dict[str, Union[List[str], Dict]], context: object) -> Dict[str,
             input (list): event['input']
     """
     print(event)
-    granule_urls = event.get('input', [])
+    event_input = event.get('input')
+    # If there is no granules object, fail the workflow.
+    granules_list = event_input.get('granules')
```
AndrewDorn, line 88: This code will actually run just fine even if 'granules' is missing. granules_list will be None.

tasks/copy_to_glacier/handler.py
```diff
@@ -115,42 +83,37 @@ def task(event: Dict[str, Union[List[str], Dict]], context: object) -> Dict[str,
...
-        print(f"Copied {granule_url} into glacier storage bucket {glacier_bucket}.")
-
-    final_output = list(granule_data.values())
-    return {'granules': final_output, 'input': granule_urls}
-
+    copied_file_urls = []
+
+    # Iterate through the input granules (>= 0 granules expected)
+    for granule in granules_list:
+        granuleId = granule['granuleId']
```
AndrewDorn, line 98: I'd be tempted to remove this single-use variable.

tasks/copy_to_glacier/handler.py
ifestus: Good catch. I will remove that comment.

tasks/copy_to_glacier/handler.py
```diff
@@ -115,42 +83,37 @@ def task(event: Dict[str, Union[List[str], Dict]], context: object) -> Dict[str,
...
+                                         source_key=source_filepath,
+                                         destination_bucket=glacier_bucket,
+                                         destination_key=source_filepath)
+            copied_file_urls.append(file['filename'])
+            print(f"Copied {source_filepath} into glacier storage bucket {glacier_bucket}.")
+
+    return {'granules': granules_list, 'copied_to_glacier': copied_file_urls}
 
 # handler that is provided to aws lambda
 def handler(event: Dict[str, Union[List[str], Dict]], context: object) -> Any:
```
AndrewDorn, line 119: Docs will need to be updated here and in the generated docs in readme

tasks/copy_to_glacier/test/test_handler.py
```diff
@@ -113,317 +129,56 @@ def test_task_happy_path(self):
...
+        self.assertEqual(event['input']['granules'], result['granules'])
         granules = result['granules']
         self.assertIsNotNone(granules)
-        granules = list(granules)
         self.assertEqual(1, len(granules))
         granule = granules[0]
-        self.assertEqual(filename, granule['granuleId'])
-        self.assertEqual(1, len(granule['files']))
+        self.assertEqual(4, len(granule['files']))
         file = granule['files'][0]
```
AndrewDorn, line 138: Unused?

tasks/copy_to_glacier/test/test_handler.py
```diff
@@ -113,317 +129,56 @@ def test_task_happy_path(self):
...
+                    'StorageClass': 'GLACIER',
+                    'MetadataDirective': 'COPY',
+                    'ContentType': content_type
                 }
-            }
+            ))
 
-        }
+        s3_cli.head_object.assert_has_calls(head_object_calls)
+        s3_cli.copy.assert_has_calls(copy_calls)
```
AndrewDorn, line 156: I personally like locking down the call_count as well to make sure only the calls you want are present.

tasks/copy_to_glacier/test/test_handler.py
```diff
@@ -113,317 +129,56 @@ def test_task_happy_path(self):
...
 
-        result = task(event, None)
+        expected_copied_file_urls = [ file['filename'] for file in self.event_granules['granules'][0]['files'] ]
+        self.assertEqual(expected_copied_file_urls, result['copied_to_glacier'])
```
AndrewDorn, line 159: How feasible would it be to check every property on 'file'?

tasks/copy_to_glacier/handler.py
AndrewDorn: Is the case where granules_list == None handled?

AndrewDorn: Doesn't look like there's a unit test around it either. I feel like that's a raiseable offense, but we could also consider it equal to granules_list being empty.

tasks/copy_to_glacier/test/test_handler.py
ifestus: I don't think there's much value in checking each property because file is just copied from the granules input object. It may be useful for checking that no changes were made to the object, though...

tasks/copy_to_glacier/test/test_handler.py
ifestus: :+1:

tasks/copy_to_glacier/handler.py
ifestus: :+1:

tasks/copy_to_glacier/test/test_handler.py
ifestus: I should just be able to do an `asssertEqual` on the output granule and input granule to check that.

PR #60, opened 2021-03-04
ORCA Lambda Module Updates

bhazuka:
## Summary of Changes

Updated variable names and conventions. Added comments and documentation in code for understanding. 

Addresses [ORCA-100: Reconcile ORCA and Cumulus Terraform variables](https://bugs.earthdata.nasa.gov/browse/ORCA-100)

## Changes

module/lambda/variables.tf

- Updated to use more conventionally named variables based on a Cumulus deployment.
- Notes and documentation added to better distinguish where variables originate, required, optional and other information.

module/lambda/main.tf

- Added terraform and AWS configuration sections.
- Added locals section for tags and other variables to guarantee deployment tag
- Added notes, documentation, and descriptions to better understand code
- Organized resources into a understandable groups
- Updated variable names based on changes to variables.tf
- Normalized naming convention based on convention in other modules. 

module/lambda/output.tf

- Added description and updated object call to match naming convention

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

This update is currently untested. Updates need to be added to CHANGELOG

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
    - [ ] CHANGELOG.md
    - [ ] Testing documentation
    - [ ] Development documentation
    - [ ] User documentation
    - [ ] (Additional here)
- [X] My changes generate no new warnings
    - [ ] Unit tests
    - [ ] Integration tests
- [X] New and existing unit tests pass locally with my changes
    - [X] Automated tests passing (if not fork)
- [ ] Any dependent changes have been merged and published in downstream modules
- [X] My code has passed security scanning
    - [X] Snyk
    - [X] git-secrets

Currently, ORCA terraform modules use various variable names throughout the modules that are not well defined and are defined differently between modules when they are actually the same thing.

Users should be able to clearly identify what variables are needed and understand how the variables are used. In addition, a user should understand which variables ORCA is utilizing from the Cumulus deployment and which are unique to ORCA.


modules/lambdas/main.tf
```diff
@@ -1,83 +1,154 @@
...
+## =============================================================================
+
+# copy_to_glacier - Copies files to the ORCA S3 Glacier bucket
+# ==============================================================================
+resource "aws_lambda_function" "copy_to_glacier" {
+  ## REQUIRED
+  function_name = "${var.prefix}_copy_to_glacier"
   role          = module.restore_object_arn.restore_object_role_arn
-  handler       = "db_deploy.handler"
-  runtime       = "python3.7"
```
rizbi-hassan, line 29: I get the following error while deploying without these two parameters:

"Error: **handler** and **runtime** must be set when PackageType is Zip"

Wondering if these two parameters should stay at ## REQUIRED parameter section in case of .zip package files

modules/lambdas/main.tf
```diff
@@ -1,83 +1,154 @@
...
-  timeout       = var.lambda_timeout
-  description   = "Deploys the Disaster Recovery database"
+
+  ## OPTIONAL
+  description      = "ORCA archiving lambda used to copy data to an ORCA S3 glacier bucket."
+  filename         = "${path.module}/../../tasks/copy_to_glacier/copy_to_glacier.zip"
+  handler          = "handler.handler"
```
rizbi-hassan, line 85: Should this be changed to "copy_to_glacier.handler"  in line 85?
The handler.py should also be changed to copy_to_glacier.py in that case.

bhazuka: No, handler.handler is correct in this context. If you look at the task https://github.com/nasa/cumulus-orca/tree/master/tasks/copy_to_glacier you will notice that the main python script is called `handler.py` and the function inside being called is `handler`. That is where the handler is being called `handler.handler`.

I agree that we should change it to better match convention from `handler.py` to `copy_to_glacier.py`. I will make those changes shortly.

modules/lambdas/main.tf
bhazuka: I based required and optional for the Lambda off of the Terraform documentation https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#argument-reference

I guess I would rather keep them under optional to keep it consistent since we may move to other methods in the future. Otherwise, I would move almost everything under required since I would expect us to use most of the fields provided with only environment being optional between lambdas. Thoughts?

rizbi-hassan: yes sounds good to me. We can keep them as it is now.

modules/lambdas/main.tf
bhazuka: Changes to lambda have been made. please review them for any additional issues.

rizbi-hassan: looks good to me.

PR #62, opened 2021-03-10
ORCA RDS Module Updates

bhazuka:
## Summary of Changes

Updated variable names and conventions. Added comments and documentation in code for understanding. 

Addresses [ORCA-100: Reconcile ORCA and Cumulus Terraform variables](https://bugs.earthdata.nasa.gov/browse/ORCA-100)

## Changes

module/rds/variables.tf

- Updated to use more conventionally named variables based on a Cumulus deployment.
- Notes and documentation added to better distinguish where variables originate, required, optional and other information.

module/rds/main.tf

- Added terraform and AWS configuration sections.
- Added locals section for tags and other variables to guarantee deployment tag
- Added notes, documentation, and descriptions to better understand code
- Organized resources into a understandable groups
- Updated variable names based on changes to variables.tf
- Normalized naming convention based on convention in other modules. 

module/rds/output.tf

- Added descriptions to output after investigating usages.

modules/lambda/output.tf

- Added additional output variable that is needed by rds module.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

This update is currently untested. Updates need to be added to CHANGELOG

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
    - [ ] CHANGELOG.md
    - [ ] Testing documentation
    - [ ] Development documentation
    - [ ] User documentation
    - [ ] (Additional here)
- [X] My changes generate no new warnings
    - [ ] Unit tests
    - [ ] Integration tests
- [X] New and existing unit tests pass locally with my changes
    - [X] Automated tests passing (if not fork)
- [ ] Any dependent changes have been merged and published in downstream modules
- [X] My code has passed security scanning
    - [X] Snyk
    - [X] git-secrets

Currently, ORCA terraform modules use various variable names throughout the modules that are not well defined and are defined differently between modules when they are actually the same thing.

Users should be able to clearly identify what variables are needed and understand how the variables are used. In addition, a user should understand which variables ORCA is utilizing from the Cumulus deployment and which are unique to ORCA.

### DIFF
--- a/modules/rds/main.tf
+++ b/modules/rds/main.tf
@@ -1,84 +1,172 @@
-resource "aws_db_subnet_group" "postgres_subnet_group" {
-  name       = "${var.prefix}-postgres-subnet-group"
-  subnet_ids = var.subnet_ids
+## Terraform Requirements
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 3.5.0"
+    }
+  }
 }
 
-# Ignore aws profile in case this was deployed with CI or
-# in a machine without aws profile defined
+
+## AWS Provider Settings
+provider "aws" {
+  region  = var.region
+  profile = var.aws_profile
+}
+
+
+## Local Variables
 locals {
-  used_profile = contains([var.profile], "default") ? "" : "--profile ${var.profile}"
-}
-# https://blog.faraday.io/how-to-create-an-rds-instance-with-terraform/
-resource "aws_db_instance" "postgresql" {
-  # set apply_immediately true just for testing
-  apply_immediately = false
-  allocated_storage          = 20
-  # max_allocated_storage      = 21
-  engine                     = "postgres"
-  engine_version             = "11"
-  identifier                 = "${var.prefix}-postgres-sndbx"
-  # Can't do micro and storage_encrypted
-  instance_class             = "db.t2.micro"
-  name                       = "postgres"
-  password                   = var.postgres_user_pw
-  username                   = "postgres"
-  iam_database_authentication_enabled = "true"
-  backup_retention_period    = 7
-  backup_window              = "03:00-06:00"
-  maintenance_window         = "Mon:00:00-Mon:03:00"
-  # auto_minor_version_upgrade = "${var.auto_minor_version_upgrade}"
-  port                       = var.database_port
-  vpc_security_group_ids     = [var.vpc_postgres_ingress_all_egress_id]
-  db_subnet_group_name       = aws_db_subnet_group.postgres_subnet_group.id
-  # parameter_group_name       = "default.postgres11"
-  storage_encrypted          = false
-  deletion_protection        = false
-  skip_final_snapshot        = true
-  final_snapshot_identifier  = "daacx"
-  enabled_cloudwatch_logs_exports = ["postgresql", "upgrade"]
+  tags = merge(var.tags, { Deployment = var.prefix })
+  # Ignore aws profile in case this was deployed with CI or in a machine without
+  # aws profile defined
+  used_profile = contains([var.aws_profile], "default") ? "" : "--profile ${var.aws_profile}"
 }
 
+
+## Resources
+
+## =============================================================================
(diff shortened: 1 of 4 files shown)


modules/rds/main.tf
```diff
@@ -1,84 +1,172 @@
...
+# https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance
+# https://registry.terraform.io/modules/terraform-aws-modules/rds/aws/latest
 resource "aws_db_instance" "postgresql" {
+  ## REQUIRED
+  allocated_storage = 20
+  engine            = "postgres"
+  identifier        = "${var.prefix}-postgres-sndbx"
+  instance_class    = "db.t3.micro"
+  password          = var.postgres_user_pw
+  username          = "postgres"
```
rizbi-hassan, line 59: should we move some of these parameters to variables.tf and then use var.xx  here? Just a thought.

modules/rds/main.tf
```diff
@@ -1,84 +1,172 @@
...
+  backup_window                       = "03:00-06:00"
+  db_subnet_group_name                = aws_db_subnet_group.postgres_subnet_group.id
+  delete_automated_backups            = true
+  deletion_protection                 = false
+  enabled_cloudwatch_logs_exports     = ["postgresql", "upgrade"]
+  engine_version                      = "11"
+  final_snapshot_identifier           = "daacx"
+  iam_database_authentication_enabled = true
+  maintenance_window                  = "Mon:00:00-Mon:03:00"
+  name                                = "postgres"
```
rizbi-hassan, line 75: maybe we can also move some of these parameters like backup_retention_period, engine_version, final_snapshot_identifier to variables.tf 
Thoughts?

modules/rds/main.tf
```diff
@@ -1,84 +1,172 @@
...
+## drdb-user-pass - Application user database password
+## =============================================================================
 resource "aws_secretsmanager_secret" "drdb-user-pass" {
-  name = "${var.prefix}-drdb-user-pass"
-  description = "User password to be used for the PostgreSQL DB"
-  tags        = var.default_tags
+  ## OPTIONAL
+  description             = "Application user password to be used for the PostgreSQL DB"
+  name                    = "${var.prefix}-drdb-user-pass"
+  recovery_window_in_days = 0
```
rizbi-hassan, line 141: also same for recovery_window_in_days. How about we move it to variables.tf?

modules/rds/main.tf
bhazuka: So the short answer is probably not as changing anything but the instance_class can have some negative impacts to our current code base. For the most part I kept the resource the same and just documented items. The one change I did make was moving from a db.t2.micro to a db.t3.micro. The reason I did this was the cost was the same and I could enable encryption which is not available with the db.t2.micro.

Realistically ORCA in it's current form is pretty light on the database and anything bigger than a micro is wasted. In addition, in the next PI (only a few weeks away) we will be working on moving over the the Cumulus RDS instance so this resource will likely be removed completely from our terraform module since we will rely on their instance. Also, LP and PO DAAC have been using an incarnation of this database for over a year with no issues and no requirements to expose those items.

That being said, if we were going to maintain our own database instance long term, I would surface items like instance_class. If you feel strongly about surfacing that variable for the short term let me know.

modules/rds/main.tf
bhazuka: I would like to keep recovery_window_in_days set to 0. This value was unset in the original and the 30 day window causes issues especially if you want to tear down your stack and re-deploy. In the original, if you tore down your stack, you had to manually go remove secrets through the dashboard which allowed you to go from a 30 day to a 7 day wait time. I don't feel this is something we should bubble up to the users but should just set for them.

modules/rds/main.tf
rizbi-hassan: yes we can then keep them in the main.tf for now then since our customers don't want that and since the RDS instance will go away.

PR #85, opened 2021-05-02
Feature/orca 165

rizbi-hassan:
## Summary of Changes

Addresses [ORCA-165: Complete post_copy_request_to_queue](https://bugs.earthdata.nasa.gov/browse/ORCA-165)

## Changes
* Updated /modules/lambda and /modules/orca to include the new lambda terraform code
* created new dir post_copy_request_to_queue under /tasks.
  * added post_copy_request_to_queue.py file
  * added build.sh and run_tests.sh shell scripts under bin folder.
  * added input.json under schemas folder
  * added README.md file
  * added requirements-dev.txt file

## Type of change

## Checklist:

- [ ] I have performed a self-review of my own code
    - [x] CHANGELOG.md


tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,99 @@
...
+    for record in event['Records']:
+        filename = record['s3']['object']['key']
+
+    sql = """
+        SELECT
+            job_id, granule_id, restore_destination
+        FROM
+            orca_recoverfile
+        WHERE
+            filename = %s
```
bhazuka, line 55: I will have to double check but I believe you may want to use key_path instead of filename because I believe the return schema ill return the full key path not just the filename.

bhazuka (changes requested):
Overall pretty good, couple of things to adjust and think about. In terraform, we need to be able to handle the potential for multiple ORCA archive (glacier) buckets.  In your code you need to be able to handle multiple rows returned. In addition, should have some error catching.

modules/lambdas/main.tf
```diff
@@ -303,6 +280,45 @@ resource "aws_lambda_function" "request_status_for_job" {
...
+# post_copy_request_to_queue -Posts to two queues for notifying copy_files_to_archive lambda and updating the DB."
+# ==============================================================================
+resource "aws_lambda_function" "post_copy_request_to_queue" {
+  ## REQUIRED
+  function_name = "${var.prefix}_post_copy_request_to_queue"
+  role          = module.restore_object_arn.restore_object_role_arn
+  ## OPTIONAL
+  description      = "Posts to two queues for notifying copy_files_to_archive lambda and updating the DB."
+  filename         = "${path.module}/../../tasks/post_copy_request_to_queue/post_copy_request_to_queue.zip"
+  handler          = "post_copy_request_to_queue.handler"
```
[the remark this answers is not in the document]
rizbi-hassan: I am a bit confused on this one. What should be changed?

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,99 @@
...
+    if len(rows) == 0:
+        return None
+    orca_recoveryfile = rows[0]
+    # grab the parameters from the db in json format
+    job_id = database.result_to_json(orca_recoveryjob)['job_id']
+    granule_id = database.result_to_json(orca_recoveryjob)['granule_id']
+    restore_destination = database.result_to_json(orca_recoveryjob)['restore_destination']
+
+    #post to staged-recovery-queue
+    shared_recovery.post_status_for_file_to_queue(
```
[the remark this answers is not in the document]
AndrewDorn: There's a PR linked in the task that had some example code for this that I believe works.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: it will return the name of the object in the s3 bucket. 
For example, if the object name in S3 that was restored is "f1.doc", it can be retrieved from the json event input like this:

        for record in event['Records']:
          filename = record['s3']['object']['key']

filename will be equal to f1.doc in this case.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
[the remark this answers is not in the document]
rizbi-hassan: I am a bit confused about this one.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -10,41 +10,43 @@
...
-DB_HOST = 'dev-drdb-host'
-DATABASE_NAME = 'postgres'
-DATABASE_PORT = '5432'
-DATABASE_USER = 'rhassan'
-OS_ENVIRON_DB_QUEUE_URL_KEY = 'DB_QUEUE_URL'
-OS_ENVIRON_RECOVERY_QUEUE_URL_KEY = 'DB_QUEUE_URL_RECOVERY'
+# instantiate CumulusLogger
+LOGGER = CumulusLogger()
 
+PREFIX = os.environ["PREFIX"]
```
bhazuka, line 27: Generally, I would not make these variables global. In reality you do not need to set the following variables in your code since the `db_connect_info = get_dbconnect_info()` does this for you.
```
PREFIX
DATABASE_NAME
DATABASE_PORT
DATABASE_USER
```

The remaining two variables DB_QUEUE_URL and RECOVERY_QUEUE_URL I would get values for in the handler and perform proper checking to make sure they are set. This includes logging and adding a raise Exception event if the values are not present or of 0 length.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -10,41 +10,43 @@
...
+DATABASE_NAME = os.environ["DATABASE_NAME"]
+DATABASE_PORT = os.environ["DATABASE_PORT"]
+DATABASE_USER = os.environ["DATABASE_USER"]
+DB_QUEUE_URL = os.environ["DB_QUEUE_URL"]
+RECOVERY_QUEUE_URL = os.environ["RECOVERY_QUEUE_URL"]
 
-def lambda_handler(event, context):
+
+def task(records):
```
bhazuka, line 35: Add type hints. See the style guide https://google.github.io/styleguide/pyguide.html#s3.8.3-functions-and-methods. You also may need to add more arguments. See the comment about the global variables.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -56,44 +58,75 @@ def lambda_handler(event, context):
...
+    db_connect_info = get_dbconnect_info()
 
     try:
-        rows = database.single_query(sql, database.db_connect_info, (filename, shared_recovery.OrcaStatus.PENDING))
-    except database.DbError as err:
-        LOGGER.error(f"DbError: {str(err)}")
-        raise DatabaseError(str(err))
+        rows = single_query(sql, db_connect_info, (filename, OrcaStatus.PENDING.value))
+    except Exception as ex:
+        LOGGER.error(ex)
```
bhazuka, line 68: I would improve the logging message here instead of just passing it the exception and stack trace. Possibly something like "Unable to retrieve {filename} metadata.". You will also want to re-raise the error here so that the code stops.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -56,44 +58,75 @@ def lambda_handler(event, context):
...
 
     if len(rows) == 0:
```
bhazuka, line 70: I would put this logic in your try block.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -56,44 +58,75 @@ def lambda_handler(event, context):
...
-        job_id = job_id,
-        granule_id = granule_id,
-        filename = filename,
-        restore_destination = restore_destination,
-        status_id = shared_recovery.OrcaStatus.STAGED,
-        error_message = '',
-        request_method = shared_recovery.RequestMethod.NEW,
-        db_queue_url= OS_ENVIRON_RECOVERY_QUEUE_URL_KEY
+        raise Exception(
+            "length of rows cannot be empty since there will always be a record in the database"
```
bhazuka, line 72: Change this message, possible something like "No metadata found for {filename}". I would also add a critical log message as well before the raise.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -56,44 +58,75 @@ def lambda_handler(event, context):
...
         )
+    # convert db result to json
+    db_result_json = result_to_json(rows)
```
bhazuka, line 75: Add this to the try block

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -56,44 +58,75 @@ def lambda_handler(event, context):
...
+
+    # grab the parameters from the db in json format
+    job_id = db_result_json["job_id"]
+    granule_id = db_result_json["granule_id"]
+    restore_destination = db_result_json["restore_destination"]
+
+    # post to staged-recovery-queue
+    post_status_for_file_to_queue(
```
bhazuka, line 83: I don't think this is the correct call. This is for updating status in the database QUEUE. I believe you would want to use the base `post_entry_to_queue` call so that you could send the proper input needed for the copy to archive lambda. See:
- https://github.com/nasa/cumulus-orca/blob/develop/tasks/copy_files_to_archive/schemas/input.json
- https://github.com/nasa/cumulus-orca/blob/develop/tasks/copy_files_to_archive/schemas/sub_schemas/body.json

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -56,44 +58,75 @@ def lambda_handler(event, context):
...
+        job_id,
+        granule_id,
+        filename,
+        restore_destination,
+        OrcaStatus.STAGED.value,
+        "",
```
bhazuka, line 89: use None instead of an empty string

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -56,44 +58,75 @@ def lambda_handler(event, context):
...
-        )
\ No newline at end of file
+    # post to DB-queue
+    post_status_for_file_to_queue(
+        job_id,
+        granule_id,
+        filename,
+        restore_destination,
+        OrcaStatus.STAGED.value,
+        "",
```
bhazuka, line 101: use None instead of an empty string

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -56,44 +58,75 @@ def lambda_handler(event, context):
...
         )
+    # convert db result to json
+    db_result_json = result_to_json(rows)
+
+    # grab the parameters from the db in json format
+    job_id = db_result_json["job_id"]
+    granule_id = db_result_json["granule_id"]
+    restore_destination = db_result_json["restore_destination"]
+
+    # post to staged-recovery-queue
```
bhazuka, line 82: You may want to put this in a try block to catch exceptions, log them and retry a certain number of times.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -56,44 +58,75 @@ def lambda_handler(event, context):
...
-        granule_id = granule_id,
-        filename = filename,
-        restore_destination = restore_destination,
-        status_id = shared_recovery.OrcaStatus.STAGED,
-        error_message = '',
-        request_method = shared_recovery.RequestMethod.NEW,
-        db_queue_url= OS_ENVIRON_DB_QUEUE_URL_KEY
-        )
\ No newline at end of file
+    # post to DB-queue
```
bhazuka, line 94: You may want to put this in a try block to catch exceptions, log them and retry a certain number of times.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -10,41 +10,43 @@
...
 from typing import Dict, Any, List, Optional
 from datetime import datetime, timezone
-import shared_recovery
-import requests_db
-import database
+from shared_recovery import (
+    OrcaStatus,
+    RequestMethod,
+    post_status_for_file_to_queue,
+    post_status_for_job_to_queue,
```
bhazuka, line 17: This library is not called, remove from import.

modules/lambdas/variables.tf
```diff
@@ -151,16 +151,28 @@ variable "orca_recovery_retrieval_type" {
...
+  type        = string
+  description = "SQS URL of recovery queue."
 }
 
 variable "platform" {
   type        = string
   description = "String that determines deployment platform. AWS or local."
+}
+
+# list of orca bucket arn needed to provide permission to post_copy_request_to_queue lambda
```
bhazuka, line 173: I would remove this variable as it is not needed.

modules/lambdas/main.tf
```diff
@@ -299,25 +299,44 @@ resource "aws_lambda_function" "post_copy_request_to_queue" {
...
+      DATABASE_PORT      = var.database_port
+      DATABASE_NAME      = var.database_name
+      DATABASE_USER      = var.database_app_user
+      DB_QUEUE_URL       = var.orca_sqs_status_update_queue_id
+      RECOVERY_QUEUE_URL = var.orca_sqs_recovery_queue_id
+    }
+  }
 }
 
 # Permissions to allow S3 trigger to invoke lambda
```
bhazuka, line 314: We should be able to construct the arns so no need to have them in a variable. see below. Also may want to change allow_bucket to something more informative like "allow_s3_trigger"

```
resource "aws_lambda_permission" "allow_s3_trigger" {
  for_each = toset(local.orca_buckets)
  statement_id  = "AllowExecutionFromS3"
  action        = "lambda:InvokeFunction"
  function_name = aws_lambda_function.post_copy_request_to_queue.function_name
  principal     = "s3.amazonaws.com"
  source_arn    = "arn:aws:s3:::${each.value}"
}
```

Note that the arn list could also be created in the local block and referenced here as another option.

modules/lambdas/variables.tf
```diff
@@ -151,16 +151,28 @@ variable "orca_recovery_retrieval_type" {
...
   description = "The ARN of the staged-recovery-queue SQS"
 }
 
 variable "orca_sqs_status_update_queue_id" {
   type        = string
-  description = "The URL of the SQS queue that recoery status updates are read from/posted to."
+  description = "The URL of the SQS queue that recovery status updates are read from/posted to."
+}
+
+variable "orca_sqs_recovery_queue_id" {
```
bhazuka, line 163: Make sure this new variable gets bubbled up to the orca/main.tf, orca/variables.tf and the main.tf and variables.tf files if needed.

bhazuka (changes requested):
Getting closer, a few follow on thing s to look at and some thoughts on caching/handling errors

tasks/post_copy_request_to_queue/schemas/input.json
```diff
@@ -1,38 +1,152 @@
...
+              ]
+            }
           },
-          "object": {
-            "key": "f1.doc",
-            "size": 1305107,
-            "eTag": "b21b84d653bb07b05b1e6b33684dc11b",
-            "sequencer": "0C0F6F405D6ED209E1"
-          }
+          "required": [
```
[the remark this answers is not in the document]
rizbi-hassan: yes I agree. the lambda will only need the filename that can be retrieved from 

records["s3"]["object"]["key"]

So I will remove all other json code expect the s3 object

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
bhazuka: Ture but does that equate to the filename in the database. If I recall, filename in the database is just the name of the file. Where you would really need the key_path out of the database to compared to the filename retrieved from the JSON record.

modules/lambdas/main.tf
rizbi-hassan: interesting! I didn't know we can use that way. Good thing to learn

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: According to the above schemas, these are the required fields needed:

`{
  "job_id": "ABCDEFGHIJKLMNOPQRSTUVWXYZ",
  "granule_id": "ABCDEFGHIJKLMNOPQRSTUVWXYZ",
  "filename": "ABCDEFGHIJ",
  "source_key": "ABCDEFGHIJK",
  "target_key": "ABCDEFGHIJKLMNOPQRSTUV",
  "restore_destination": "ABCDEFGHIJKLMNOPQRSTUVWXYZABC",
  "source_bucket": "ABCDEFGHIJKLMNOPQ"
}`

How to get the source_key, target_key? I have not seen them anywhere before.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -37,104 +45,119 @@ def task(records: Dict[str, Any], db_queue_url: str, recovery_queue_url: str) ->
...
+            my_base_delay = exponential_delay(my_base_delay, copy_retry_backoff)
             continue
         break
     else:
         raise Exception("Unable to send message to recovery SQS.")
 
-    # post to DB-queue. Retry 3 times if it fails
-    for retry in range(3):
+    # post to DB-queue. Retry using exponential delay if it fails
+    for retry in range(max_retries):
```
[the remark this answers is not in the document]
rizbi-hassan: that should be copy_retries

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -37,104 +45,119 @@ def task(records: Dict[str, Any], db_queue_url: str, recovery_queue_url: str) ->
...
                 db_queue_url,
             )
         except Exception:
-            LOGGER.error("Unable to send message to db SQS. Retrying...")
+            LOGGER.error(f"Ran into error {retry+1} time(s)")
+            my_base_delay = exponential_delay(my_base_delay, copy_retry_backoff)
             continue
         break
     else:
         raise Exception("Unable to send message to db SQS.")
```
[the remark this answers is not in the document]
rizbi-hassan: added 
`LOGGER.critical(f"unable to send message to {db_queue_url} for job_id: {job_id}, granule_id: {granule_id}, filename: {filename}, status: {OrcaStatus.STAGED.name}")`

tasks/post_copy_request_to_queue/test/unit_tests/test_post_copy_request_to_queue.py
```diff
@@ -1 +1,318 @@
...
+from requests_db import get_dbconnect_info, DatabaseError
+from database import single_query, result_to_json
+from botocore.exceptions import ClientError
+
+class TestPostCopyRequestToQueue(TestCase):
+    """
+    Unit tests for the post_copy_request_to_queue lambda function.
+    """
+    # Create the mock instance of secrets manager
+    mock_sm = mock_secretsmanager()
```
bhazuka, line 23: Not sure why you are mocking secrets manager here. You would probably want to mock SQS. 

Looking at your setup, you would also want to mock `get_dbconnect_info` in the tests that require it.

rizbi-hassan: removed secrets manager and added mock get_dbconnect_info

tasks/post_copy_request_to_queue/test/unit_tests/test_post_copy_request_to_queue.py
```diff
@@ -0,0 +1,194 @@
...
+            "DATABASE_PORT": "5432",
+            "DATABASE_NAME": "disaster_recovery",
+            "DATABASE_USER": "rhassan",
+        },
+        clear=True,
+    )
+    @patch("database.single_query")
+    @patch("requests_db.get_dbconnect_info")
+
+    def test_task_happy_path(self, mock_get_db_connect_info: MagicMock, mock_single_query: MagicMock):
```
AndrewDorn, line 72: Add in mocks+asserts on post_file_to_queue

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,228 @@
...
+    Returns:
+        None
+    Raises:
+        Exception: If unable to retrieve key_path or db parameters, convert db result to json,
+        or post to queue.
+
+    """
+    # grab the key_path and bucketname from event
+    # We only expect one record.
+    for record in records:
```
AndrewDorn, line 61: This loop is deceptive, since it implies that multiple values are okay. Maybe we want to extract this out to another method? Something like
task(records, ...):
    for record in records:
        inner_task(record, ...)

inner_task(record, ...):
    ...

At minimum, remove the loop and have the code raise an error if more than one item is present.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,228 @@
...
+        retry_backoff: The multiplier by which the retry interval increases during each attempt.
+    Returns:
+        None
+    Raises:
+        Exception: If unable to retrieve key_path or db parameters, convert db result to json,
+        or post to queue.
+
+    """
+    # grab the key_path and bucketname from event
+    # We only expect one record.
```
AndrewDorn, line 60: Is there a value in the aws_s3_bucket_notification that gives us only one value in a list?

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,228 @@
...
+import database
+from cumulus_logger import CumulusLogger
+import logging
+
+# instantiate CumulusLogger
+LOGGER = CumulusLogger()
+
+
+def task(
+    records: Dict[str, Any],
```
AndrewDorn, line 31: Code suggests that this will be a List[Dict[str, Any]]

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,228 @@
...
+            message = f"No metadata found for {key_path}"
+            LOGGER.fatal(message)
+            raise Exception(message)
+        # # convert db result to json
+        db_result_json = database.result_to_json(rows)
+
+    except Exception as ex:
+        message = f"Unable to retrieve {key_path} metadata"
+        LOGGER.error(message)
+        raise ex(message)
```
AndrewDorn, line 90: This threw an error when your test code triggered it. ex is non-callable.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,228 @@
...
+        "job_id": job_id,
+        "granule_id": granule_id,
+        "filename": filename,
+        "source_key": key_path,
+        "target_key": key_path,  # todo add a card to configure target_key in the future
+        "restore_destination": restore_destination,
+        "source_bucket": bucket_name,
+    }
+    # post to recovery queue. Retry using exponential delay if it fails
+    for retry in range(max_retries):
```
AndrewDorn, line 110: For max_retries = 1, this will only run once, rather than twice.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,228 @@
...
+        break
+    else:
+        message = f"Error sending message to {recovery_queue_url} for {new_data}"
+        logging.critical(message)
+        raise Exception(message)
+
+    # resetting my_base_delay
+    my_base_delay = retry_sleep_secs
+    # post to DB-queue. Retry using exponential delay if it fails
+    for retry in range(max_retries):
```
AndrewDorn, line 130: Same range issue

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,228 @@
...
+    }
+    # post to recovery queue. Retry using exponential delay if it fails
+    for retry in range(max_retries):
+        try:
+            post_entry_to_queue(
+                "orca_recoveryfile", new_data, RequestMethod.NEW, recovery_queue_url
+            )
+        except Exception:
+            LOGGER.error(
+                f"Ran into error posting to SQS {recovery_queue_url} {retry+1} time(s)"
```
AndrewDorn, line 117: Recommend including the exception in the message.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,228 @@
...
+            )
+            my_base_delay = exponential_delay(my_base_delay, retry_backoff)
+            continue
+        break
```
AndrewDorn, line 121: If you move the break into the try, it groups the success code together.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,228 @@
...
+                RequestMethod.UPDATE,
+                db_queue_url,
+            )
+        except Exception:
+            LOGGER.error(
+                f"Ran into error posting to SQS {db_queue_url} {retry+1} time(s)"
+            )
+            my_base_delay = exponential_delay(my_base_delay, retry_backoff)
+            continue
+        break
```
AndrewDorn, line 149: If you move the break into the try, it groups the success code together.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,228 @@
...
+                key_path,
+                restore_destination,
+                OrcaStatus.STAGED,
+                None,
+                RequestMethod.UPDATE,
+                db_queue_url,
+            )
+        except Exception:
+            LOGGER.error(
+                f"Ran into error posting to SQS {db_queue_url} {retry+1} time(s)"
```
AndrewDorn, line 145: Recommend including the exception in the message.

tasks/post_copy_request_to_queue/test/unit_tests/test_post_copy_request_to_queue.py
```diff
@@ -0,0 +1,194 @@
...
+        for name in env_names:
+            value = os.getenv(name)
+            with self.subTest(
+                name=name, value=value
+            ):
+                if name in ["MAX_RETRIES", "RETRY_SLEEP_SECS", "RETRY_BACKOFF"]:
+                    self.assertRaises(ValueError, handler, self.event, context=None)
+
+    # --------------------------------------exponential delay test---------------------------------
+    def test_exponential_delay(self):
```
AndrewDorn, line 186: Recommend including something about non-integer args in test name.

AndrewDorn (changes requested):

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: we will only have one item according to our design. I have added that as a comment above

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: we will only have one item according to our design. I have added that as a comment above. Also talked with Brad regarding this

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: changed to List[Dict[str, Any]]

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: I got the exception when running from my side. maybe retry and see.

tasks/post_copy_request_to_queue/test/unit_tests/test_post_copy_request_to_queue.py
rizbi-hassan: test name changed to test_exponential_delay_non_int(self)

modules/lambdas/variables.tf
```diff
@@ -118,10 +118,6 @@ variable "orca_recovery_retry_interval" {
   description = "Number of seconds to wait between recovery failure retries."
 }
 
-variable "orca_recovery_retry_backoff" {
```
bhazuka: Did you mean to delete this? I think this is still used.

modules/lambdas/variables.tf
```diff
@@ -161,15 +144,5 @@ variable "orca_sqs_staged_recovery_queue_arn" {
 
 variable "orca_sqs_status_update_queue_id" {
   type        = string
-  description = "The URL of the SQS queue that recovery status updates are read from/posted to."
-}
-
-variable "orca_sqs_recovery_queue_id" {
```
bhazuka: Did you mean to delete this?

bhazuka (changes requested):
Terraform validate fails. make sure to merge develop into your branch to rebase and double check that you validate the terraform. This will go a long way to fix any issues before we try to do a deploy for testing.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
AndrewDorn: I get that, but if you see a loop in code is your first thought
"This code can loop over multiple items"
or
"This code only expects one item."

At minimum, recommend replacing with key_path = records[0]["s3"]["object"]["key"]

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
AndrewDorn: But what in our terraform enforces that limit? See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_event_source_mapping for the batch_size property. Apparently it defaults to 10.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
AndrewDorn: The error we saw was raised from an un-mocked get_dbconnect_info, so it is an accurate reflection of a real error.

modules/lambdas/variables.tf
rizbi-hassan: nope my bad

modules/lambdas/variables.tf
rizbi-hassan: nope my bad

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: changed to records[0]["s3"]["object"]["key"]

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: changed to records[0]["s3"]["object"]["key"]

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: current I dont have any aws secretsmanager secrets in my free tier aws account so it was giving an error of "no GetSecret". 
I think if you run it in the NGAP aws account, it should work

AndrewDorn: Right, but if an error IS thrown, then this error handling code will fail.

tasks/post_copy_request_to_queue/README.md
```diff
@@ -0,0 +1,81 @@
+[![Known Vulnerabilities](https://snyk.io/test/github/nasa/cumulus-orca/badge.svg?targetFile=tasks/post_copy_request_to_queue/requirements.txt)](https://snyk.io/test/github/nasa/cumulus-orca?targetFile=tasks/copy_files_to_archive/requirements.txt)
+
+**Lambda function post_copy_request_to_queue **
```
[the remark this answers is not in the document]
rizbi-hassan: already changed.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: if an error is thrown, it will raise an exception.

except Exception as ex:        
message = f"Unable to retrieve {key_path} metadata"        
LOGGER.error(message)        
raise ex(message)

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: so this should be for retry in range(max_retries+1) ?

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: so this should be for retry in range(max_retries+1) ?

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: will change to:

except Exception as ex:
  LOGGER.error(
   f"Ran into error posting to SQS {recovery_queue_url} {retry+1} time(s) with exception {ex}"
)

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: will change to:

except Exception as ex:
LOGGER.error(
f"Ran into error posting to SQS {db_queue_url} {retry+1} time(s) with exception {ex}"
)

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: I think we can keep the break like this. I have not seen any error using this though

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: I think we can keep the break like this. I have not seen any error using this though

rizbi-hassan:
New PR is sent at https://github.com/nasa/cumulus-orca/pull/93

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
AndrewDorn: But if the batch size defaults to 10, then this will only process the first item. Please explain to me how you know that only one item at a time will enter this lambda.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
AndrewDorn: Nope. Run this code.

try:
    raise Exception("blah")
except Exception as ex:
    raise ex("hello")

You will get the error
TypeError: 'Exception' object is not callable

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
AndrewDorn: Yes

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: we should change the [batch_size](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_event_source_mapping#batch_size) parameter for resource "aws_lambda_event_source_mapping to 1 instead of 10 in terraform config. Then only one item will enter lambda at a time.

The batch_size is only applicable when the message goes from SQS-> lambda. In post_copy_request_to_queue, that is not the case. We should change this in terraform config of copy_files_to_archive lambda

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: ok that makes sense. yes you are right. Changed to the following: 

try:
raise Exception("blah")
except Exception:
raise Exception("hello")

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
AndrewDorn: Yes

PR #93, opened 2021-05-24
Feature/orca 165

rizbi-hassan:
## Summary of Changes

This is a new PR being made due to some issues with my previous forked repo getting deleted.
This PR is basically the copy of this PR https://github.com/nasa/cumulus-orca/pull/85 with some minor fixes

Addresses [ORCA-165: Complete post_copy_request_to_queue](https://bugs.earthdata.nasa.gov/browse/ORCA-165)
## Changes
* Updated /modules/lambda and /modules/orca to include the new lambda terraform code
* created new dir post_copy_request_to_queue under /tasks.
  * added post_copy_request_to_queue.py file
  * added test/unit_tests/test_post_copy_request_to_queue.py unit tests
  * added build.sh and run_tests.sh shell scripts under bin folder.
  * added input.json under schemas folder
  * added README.md file
  * added requirements-dev.txt file

## Type of change

## Checklist:

- [ ] I have performed a self-review of my own code
    - [x] CHANGELOG.md


tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,218 @@
...
+        retry_backoff: The multiplier by which the retry interval increases during each attempt.
+    Returns:
+        None
+    Raises:
+        Exception: If unable to retrieve key_path or db parameters, convert db result to json,
+        or post to queue.
+
+    """
+    # grab the key_path and bucketname from event
+    # We only expect one record.
```
AndrewDorn, line 54: If the batch size defaults to 10, then this will only process the first item. Please explain to me how you know that only one item at a time will enter this lambda.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,218 @@
...
+            message = f"No metadata found for {key_path}"
+            LOGGER.fatal(message)
+            raise Exception(message)
+        # # convert db result to json
+        db_result_json = database.result_to_json(rows)
+
+    except Exception as ex:
+        message = f"Unable to retrieve {key_path} metadata"
+        LOGGER.error(message)
+        raise ex(message)
```
AndrewDorn: This still doesn't work. Run this code:

try:
raise Exception("blah")
except Exception as ex:
raise ex("hello")

You will get the error
TypeError: 'Exception' object is not callable

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,218 @@
...
+            shared_recovery.post_entry_to_queue(
+                new_data, shared_recovery.RequestMethod.NEW_JOB, recovery_queue_url
+            )
+        except Exception as ex:
+            LOGGER.error(
+                f"Ran into error posting to SQS {recovery_queue_url} {retry+1} time(s) with exception {ex}"
+            )
+            my_base_delay = exponential_delay(my_base_delay, retry_backoff)
+            continue
+        break
```
AndrewDorn, line 114: I understand that this works, but having the break here means that somebody reading your code has to jump between sections to understand a successful iteration.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,218 @@
...
+                None,
+                db_queue_url,
+            )
+        except Exception as ex:
+            LOGGER.error(
+                f"Ran into error posting to SQS {db_queue_url} {retry+1} time(s) with exception {ex}"
+            )
+            my_base_delay = exponential_delay(my_base_delay, retry_backoff)
+            continue
+        break
```
AndrewDorn, line 139: I understand that this works, but having the break here means that somebody reading your code has to jump between sections to understand a successful iteration.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,218 @@
...
+    Returns:
+        None
+    Raises:
+        Exception: If unable to retrieve key_path or db parameters, convert db result to json,
+        or post to queue.
+
+    """
+    # grab the key_path and bucketname from event
+    # We only expect one record.
+    key_path = records["s3"]["object"]["key"]
```
AndrewDorn, line 55: I saw a comment suggesting you'd changed to records[0]["s3"]. Is that on its way in?

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
AndrewDorn: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_event_source_mapping

bhazuka (changes requested):
A few minor changes and some thoughts and questions to clarify a few items. Let me know what you think or if you have questions.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
AndrewDorn: On further reading, it looks like you're now calling task with a single dict, but haven't updating naming or docs.

AndrewDorn (changes requested):
I still have concerns around how/if you're enforcing a one-item-at-a-time policy for the queue.

tasks/post_copy_request_to_queue/README.md
```diff
@@ -0,0 +1,90 @@
+[![Known Vulnerabilities](https://snyk.io/test/github/nasa/cumulus-orca/badge.svg?targetFile=tasks/post_copy_request_to_queue/requirements.txt)](https://snyk.io/test/github/nasa/cumulus-orca?targetFile=tasks/copy_files_to_archive/requirements.txt)
+
+**Lambda function post_copy_request_to_queue **
```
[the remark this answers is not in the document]
rizbi-hassan: changed to # Lambda function post_copy_request_to_queue

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,218 @@
...
+import os
+import boto3
+from typing import Dict, Any, List
+import time
+import random
+from orca_shared import shared_recovery
+import requests_db
+import database
+from cumulus_logger import CumulusLogger
+import logging
```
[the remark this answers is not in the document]
rizbi-hassan: that is being used in 117 and 142 to log the new_data. I was having issues to log the json using CumulusLogger  but didn't get the error while using logging.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,218 @@
...
+        break
+    else:
+        message = f"Error sending message to {recovery_queue_url} for {new_data}"
+        logging.critical(message)
+        raise Exception(message)
+
+    # resetting my_base_delay
+    my_base_delay = retry_sleep_secs
+    # post to DB-queue. Retry using exponential delay if it fails
+    for retry in range(max_retries+1):
```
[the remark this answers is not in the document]
rizbi-hassan: ok changed the logic. first it will try to send to db queue and then to recovery queue.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: did you try this: 

try:
raise Exception("blah")
except Exception:
raise Exception("hello")

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: we should change the [batch_size](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_event_source_mapping#batch_size) parameter for resource "aws_lambda_event_source_mapping to 1 instead of 10 in terraform config. Then only one item will enter lambda at a time.

The batch_size is only applicable when the message goes from SQS-> lambda. In post_copy_request_to_queue, that is not the case. We should change this in terraform config of copy_files_to_archive lambda

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: check line 217
records = event["Records"][0]

also updated docstring in task def to reflect a dict instead of a list of dict.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: also the event in this case is coming from S3, not SQS. so there is no batch_size.

AndrewDorn: Between me looking at the wrong docs, and the right docs not existing, let's just add a comment linking to https://forums.aws.amazon.com/thread.jspa?messageID=592264#592264 and call it good.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,218 @@
...
+
+        if var in ["MAX_RETRIES", "RETRY_SLEEP_SECS", "RETRY_BACKOFF"]:
+            try:
+                env_var_value = int(env_var_value)
+            except ValueError as ve:
+                LOGGER.critical(f"{var} must be set to an integer.")
+                raise ve
+        backoff_args.append(env_var_value)
+
+    records = event["Records"][0]
```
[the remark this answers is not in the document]
rizbi-hassan: the event is actually coming from s3 rather than SQS.
for every object restored in s3 from glacier, it will send one event to the lambda instead of a list. 

I discussed with @AndrewDorn  regarding this. Check his [comment](https://github.com/nasa/cumulus-orca/pull/93#discussion_r638296999)

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
AndrewDorn: That's not what you're doing. Look at your line 83: "raise ex(message)"

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
AndrewDorn: You're right, allow me to be more specific... If you get a single item from records, then I wouldn't use a plural name. Might want to switch from 'records' to 'record'.

tasks/post_copy_request_to_queue/test/unit_tests/test_post_copy_request_to_queue.py
```diff
@@ -0,0 +1,342 @@
...
+                        handler(self.event, context=None)
+                        self.assertEquals(ve.message, message)
+
+    # --------------------------------------exponential delay test---------------------------------
+    def test_exponential_delay_non_int(self):
+        """
+        tests delay function. Raises TypeError when args are non-integer
+        """
+
+        base_delay = "non-integer"
```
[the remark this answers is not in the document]
rizbi-hassan: looks like this function is tested in other unit tests while retrying. So do we really need this test?

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: ok my bad. should be good now

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: moved 'break' inside 'try'

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: moved 'break' inside 'try'

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
AndrewDorn: Great. We should log errors though.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -0,0 +1,219 @@
...
+        db_connect_info = requests_db.get_dbconnect_info()
+        rows = database.single_query(sql, db_connect_info, (key_path, shared_recovery.OrcaStatus.PENDING.value))
+        if len(rows) == 0:
+            message = f"No metadata found for {key_path}"
+            LOGGER.fatal(message)
+            raise Exception(message)
+        # # convert db result to json
+        db_result_json = database.result_to_json(rows)
+
+    except Exception:
```
AndrewDorn, line 80: Should probably log the exception. In the message if nothing else.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: https://forums.aws.amazon.com/thread.jspa?messageID=592264

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: makes sense. Changed name to 'record'

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: changed to 

except Exception:
        message = f"Unable to retrieve {key_path} metadata. Got Exception: {Exception}"
        LOGGER.error(message)
        raise Exception(message)

bhazuka: I don't think `Got Exception: {Exception}` will return what you want. Are we looking for a callback trace here or a message? In its current form you would just get the object name like `<class 'Exception'>`. I would think you would want something like:

```
except Exception as e:
    message = f"Unable to retrieve {key_path} metadata. Got Exception: {e}"
    LOGGER.error(message, exc_info=True) # If you want tracing https://github.com/nasa/cumulus-message-adapter-python 
    raise Exception from e # Exception chaining https://docs.python.org/3/tutorial/errors.html#exception-chaining
```

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -99,50 +99,51 @@ def task(
...
-                f"Ran into error posting to SQS {recovery_queue_url} {retry+1} time(s) with exception {ex}"
+                f"Ran into error posting to SQS {db_queue_url} {retry+1} time(s) with exception {ex}"
             )
             my_base_delay = exponential_delay(my_base_delay, retry_backoff)
             continue
-        break
     else:
-        message = f"Error sending message to {recovery_queue_url} for {new_data}"
+        message = f"Error sending message to {db_queue_url} for {new_data}"
         logging.critical(message)
```
bhazuka, line 123: Shouldn't this be `LOGGER.critical` instead of `logging.critical`?

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -99,50 +99,51 @@ def task(
...
-                f"Ran into error posting to SQS {db_queue_url} {retry+1} time(s) with exception {ex}"
+                f"Ran into error posting to SQS {recovery_queue_url} {retry+1} time(s) with exception {ex}"
             )
             my_base_delay = exponential_delay(my_base_delay, retry_backoff)
             continue
-        break
     else:
-        message = f"Error sending message to {db_queue_url} for {new_data}"
+        message = f"Error sending message to {recovery_queue_url} for {new_data}"
         logging.critical(message)
```
bhazuka, line 144: Shouldn't this be LOGGER.critical instead of logging.critical?

bhazuka (changes requested):
We should investigate the error you are getting with CumulusLogger. You wouldn't wan't to mix the 2. If you use logging, you will need to setup the format and stream handlers that CumulusLogger does for us so that we are in a format that the metrics team may be able to use later down the road.

As for the test, let's talk through it.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: changed to using cumulus logger

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: changed to using cumulus logger

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -119,9 +118,9 @@ def task(
             my_base_delay = exponential_delay(my_base_delay, retry_backoff)
             continue
     else:
-        message = f"Error sending message to {db_queue_url} for {new_data}"
-        logging.critical(message)
-        raise Exception(message)
+        message = "Error sending message to db_queue_url for {new_data}"
+        LOGGER.critical(message, new_data = str(new_data))  #PO DAAC needs to fix cumulus logger to handle dict
```
bhazuka, line 122: Change comment from `#PO DAAC needs to fix cumulus logger to handle dict` to `#Cumulus will update this library in the future to be better behaved.` or something similar.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -140,9 +139,9 @@ def task(
             my_base_delay = exponential_delay(my_base_delay, retry_backoff)
             continue
     else:
-        message = f"Error sending message to {recovery_queue_url} for {new_data}"
-        logging.critical(message)
-        raise Exception(message)
+        message = "Error sending message to recovery_queue_url for {new_data}"
+        LOGGER.critical(message, new_data = str(new_data))  #PO DAAC needs to fix cumulus logger to handle dict
```
bhazuka, line 143: Change comment from `#PO DAAC needs to fix cumulus logger to handle dict` to `#Cumulus will update this library in the future to be better behaved.` or something similar.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -157,11 +156,14 @@ def exponential_delay(base_delay: int, exponential_backoff: int = 2) -> int:
     Raises:
         None
     """
-    delay = base_delay + (random.randint(0, 1000) / 1000.0)
-    time.sleep(delay)
-    LOGGER.debug(f"Performing back off retry sleeping {delay} seconds")
-    return base_delay * exponential_backoff
-
+    if type(base_delay) ==int and type(exponential_backoff)==int:
```
bhazuka, line 159: Change to try/except block catching a ValueError exception only.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -157,11 +156,14 @@ def exponential_delay(base_delay: int, exponential_backoff: int = 2) -> int:
...
+        delay = base_delay + (random.randint(0, 1000) / 1000.0)
+        time.sleep(delay)
+        LOGGER.debug(f"Performing back off retry sleeping {delay} seconds")
```
bhazuka, line 162: Flip log message and sleep command.

bhazuka (changes requested):
Few minor things, really shaping up.

bhazuka (commented):
Looks good to me.

bhazuka (approved):

AndrewDorn (approved):

modules/lambdas/variables.tf
```diff
@@ -136,6 +136,19 @@ variable "database_name" {
 }
 
 
+variable "ddl_dir" {
```
bhazuka, line 139: ddl_dir and drop_database variables are no longer needed and should be removed.

bhazuka (changes requested):
You have some old variables in lambda/variables.tf that should be removed.

bhazuka (approved):

PR #99, opened 2021-06-02
Feature/orca 177

AndrewDorn:
## Summary of Changes

Addresses [ORCA-177: Implement api gateway for the request_status lambdas](https://bugs.earthdata.nasa.gov/browse/ORCA-177)

## Changes

* Adds API Gateway to tf
* Modifies response code based on return value

## Type of change

## How Has This Been Tested?

Deployed into AWS with a dummy version of the lambda that simulates error and success codes.

## Checklist:

    - [x] CHANGELOG.md
    - [ ] Unit tests
    - [ ] Integration tests
    - [x] Snyk
    - [x] git-secrets

### DIFF
--- a/modules/lambdas/main.tf
+++ b/modules/lambdas/main.tf
@@ -298,6 +298,79 @@ resource "aws_lambda_function" "request_status_for_granule" {
   }
 }
 
+# API Gateway
+resource "aws_api_gateway_rest_api" "request_status_for_granule_api" {
+  name = "${var.prefix}_request_status_for_granule_api"
+}
+
+resource "aws_api_gateway_resource" "request_status_for_granule_api_resource" {
+  path_part   = "{proxy+}"
+  parent_id   = aws_api_gateway_rest_api.request_status_for_granule_api.root_resource_id
+  rest_api_id = aws_api_gateway_rest_api.request_status_for_granule_api.id
+}
+
+resource "aws_api_gateway_method" "request_status_for_granule_api_method" {
+  rest_api_id   = aws_api_gateway_rest_api.request_status_for_granule_api.id
+  resource_id   = aws_api_gateway_resource.request_status_for_granule_api_resource.id
+  http_method   = "ANY"
+  # todo: Make sure this is locked down against external access.
+  authorization = "NONE"
+}
+
+# todo: Error coes?
+resource "aws_api_gateway_integration" "request_status_for_granule_api_integration" {
+  rest_api_id             = aws_api_gateway_rest_api.request_status_for_granule_api.id
+  resource_id             = aws_api_gateway_resource.request_status_for_granule_api_resource.id
+  http_method             = aws_api_gateway_method.request_status_for_granule_api_method.http_method
+  integration_http_method = "POST"
+  type                    = "AWS"
+  uri                     = aws_lambda_function.request_status_for_granule.invoke_arn
+}
+
+resource "aws_api_gateway_method_response" "response_200" {
+  rest_api_id = aws_api_gateway_rest_api.request_status_for_granule_api.id
+  resource_id = aws_api_gateway_resource.request_status_for_granule_api_resource.id
+  http_method = aws_api_gateway_method.request_status_for_granule_api_method.http_method
+  status_code = "200"
+}
+
+resource "aws_api_gateway_integration_response" "request_status_for_granule_api_response" {
+  depends_on = [aws_api_gateway_integration.request_status_for_granule_api_integration]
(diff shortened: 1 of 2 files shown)


modules/lambdas/main.tf
```diff
@@ -298,6 +298,75 @@ resource "aws_lambda_function" "request_status_for_granule" {
...
+  path_part   = "{proxy+}"
+  parent_id   = aws_api_gateway_rest_api.request_status_for_granule_api.root_resource_id
+  rest_api_id = aws_api_gateway_rest_api.request_status_for_granule_api.id
+}
+
+resource "aws_api_gateway_method" "request_status_for_granule_api_method" {
+  rest_api_id   = aws_api_gateway_rest_api.request_status_for_granule_api.id
+  resource_id   = aws_api_gateway_resource.request_status_for_granule_api_resource.id
+  http_method   = "ANY"
+  # todo: Make sure this is locked down against external access.
```
AndrewDorn, line 316: I know we wanted to make use of Cumulus' security, but I don't know if we have a task for that or if things have to change here.

modules/lambdas/main.tf
```diff
@@ -298,6 +298,75 @@ resource "aws_lambda_function" "request_status_for_granule" {
...
+#set($context.responseOverride.status = 500)
+#elseif($input.path("httpStatus") != '')
+#set($context.responseOverride.status = $input.path('httpStatus'))
+#end
+EOF
+  }
+}
+
+# Lambda
+resource "aws_lambda_permission" "request_status_for_granule_api_permission" {
```
AndrewDorn, line 359: If we do want to lock down to Cumulus' security, the docs I've read suggest we need more here.

modules/lambdas/main.tf
bhazuka: From what I recall, our API would be private and only accessible to the Cumulus API. See the notes from https://wiki.earthdata.nasa.gov/pages/viewpage.action?pageId=202211487

I believe Cumulus will handle all the authorization so initially we should be able to have this as None.

modules/lambdas/main.tf
bhazuka: I feel this is something we definitely want to do long term for security. We will probably want to chat with the Cumulus folks about it to make sure we are all on the same page and that we grab the correct AWS API Gateway ARN that Cumulus will be calling our Gateway from. In the short term we may want to leave this simply locked to the accountID so we could call it without going through Cumulus but, it is not open to the world. I would create a card for security on the API Gateway so we can talk/plan/implement with Cumulus and then setup the security so that the source_arn is not set but the principal is accountId possibly. So at least it is locked to VPC initially. We can pull the accountID through a data structure. See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/caller_identity. This is just a thought, the main thing is in a Prod environment it would be locked down.

bhazuka (changes requested):
A few minor items. Overall looks good.

I wonder if we shouldn't create an API feature branch from develop and merge this and the other API cards into that initially since this may be a feature we work on for a while instead of going directly into develop.

Thoughts?

modules/lambdas/main.tf
```diff
@@ -298,6 +298,75 @@ resource "aws_lambda_function" "request_status_for_granule" {
...
+  http_method = aws_api_gateway_method.request_status_for_granule_api_method.http_method
+  status_code = aws_api_gateway_method_response.response_200.status_code
+
+  # Transforms the backend JSON response to XML
+  response_templates = {
+    "application/json" = <<EOF
+#set($inputRoot = $input.path('$'))
+$input.json("$")
+
+#if($input.path("stackTrace") != '')
```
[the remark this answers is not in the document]
AndrewDorn: Oh, that's just how it's written. Super weird, but it runs.

modules/lambdas/main.tf
AndrewDorn: Hm, it didn't like that. Said it gave a permission error. We'll stick with this for the time being.

AndrewDorn: https://bugs.earthdata.nasa.gov/browse/ORCA-206

modules/lambdas/main.tf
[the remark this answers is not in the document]
bhazuka: Per @AndrewDorn the language may be https://velocity.apache.org/engine/devel/vtl-reference.html#ifelseifelse-output-conditional-on-truth-of-statements

Quarriers89:
Looks good to me, Andrew.

PR #102, opened 2021-06-08
Added API gateway for orca_catalog_reporting

AndrewDorn:
## Summary of Changes

Addresses [ORCA-184: Implement api gateway for the reconciliation lambdas.](https://bugs.earthdata.nasa.gov/browse/ORCA-184)

## Changes

* Added API gateway for reconciliation lambda.

## Type of change

## How Has This Been Tested?

Deployed to AWS and run manually.

## Checklist:

    - [x] CHANGELOG.md
    - [x] Snyk
    - [x] git-secrets

### DIFF
--- a/modules/lambdas/main.tf
+++ b/modules/lambdas/main.tf
@@ -355,7 +355,6 @@ EOF
   }
 }
 
-# Lambda
 resource "aws_lambda_permission" "request_status_for_granule_api_permission" {
   statement_id  = "AllowExecutionFromAPIGateway"
   action        = "lambda:InvokeFunction"
@@ -456,7 +455,6 @@ EOF
   }
 }
 
-# Lambda
 resource "aws_lambda_permission" "request_status_for_job_api_permission" {
   statement_id  = "AllowExecutionFromAPIGateway"
   action        = "lambda:InvokeFunction"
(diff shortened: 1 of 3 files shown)


modules/lambdas/main.tf
```diff
@@ -552,6 +550,73 @@ resource "aws_lambda_function" "orca_catalog_reporting_dummy" {
...
+#set($context.responseOverride.status = $input.path('httpStatus'))
+#end
+EOF
+  }
+}
+
+resource "aws_lambda_permission" "orca_catalog_reporting_api_permission" {
+  statement_id  = "AllowExecutionFromAPIGateway"
+  action        = "lambda:InvokeFunction"
+  function_name = aws_lambda_function.orca_catalog_reporting_dummy.function_name
```
rizbi-hassan, line 613: Maybe move the parameters under resource to ##OPTIONAL and #REQUIRED sections. See lambda resources for example.

rizbi-hassan (changes requested):

modules/lambdas/main.tf
AndrewDorn: I would honestly prefer it if anybody looking to remove/modify these parameters looked at the docs... Since optional params have defaults, I don't think grouping them saves anything.

AndrewDorn: I guess the one time it helps is to identify variables that can definitely NOT be removed? I don't like that it implies that the other variables can be.

rizbi-hassan: ya I think moving the parameters under those two sections will help others to know which one cannot be removed.

AndrewDorn: But if you remove any of these, the code won't function as intended... You'll either get a blatant error, or the default value will break something. I'm having trouble thinking of a use case where somebody looks at the tf, sees that some variables are "optional" vs "required", and that informs a decision.

PR #107, opened 2021-06-22
Feature/orca 203

rizbi-hassan:
## Summary of Changes

Addresses [ORCA-203: Shared recovery library needs AWS Region to be variable](https://bugs.earthdata.nasa.gov/browse/ORCA-203)

## Changes

* updated tasks/shared_libraries/recovery/shared_recovery.py by adding a function that gets the aws region.
* updated tasks/shared_libraries/recovery/README.md  by adding that terraform config must include the aws region variable.
* updated /modules/lambdas/main.tf by adding a new variable "AWS_REGION_VAR" to the environment variable section of the following lambdas that use the shared lib
    * request_files
    * post_copy_request_to_queue
    * copy_files_to_archive
    * post_to_database


tasks/shared_libraries/recovery/README.md
```diff
@@ -60,22 +60,24 @@ the repository.
 
 To use the **shared_recovery** library in your lambda code base perform the following.
 
-1. Install the necessary python libraries shown below.
+1. Keep in mind that the shared library is dependent upon the AWS environment variable "AWS_REGION" and will throw an error if this variable is missing.
```
bhazuka, line 63: I would add some verbiage that says this is an AWS reserved runtime variable and maybe a link to the documentation. https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html

tasks/shared_libraries/recovery/shared_recovery.py
```diff
@@ -7,7 +7,12 @@
 import boto3
 from typing import Dict, Any, Optional, List
 from datetime import datetime, timezone
+import os
+import boto3
```
bhazuka, line 11: boto3 imported twice. Once in line 7 and then again here.

tasks/shared_libraries/recovery/shared_recovery.py
```diff
@@ -28,6 +33,22 @@ class OrcaStatus(Enum):
...
+        Raises:
+            Exception: Thrown if AWS region is empty or None.
+    """
+    logger.debug("Getting environment variable AWS_REGION value.")
+    aws_region = os.getenv("AWS_REGION", None)
+    if aws_region is None or len(aws_region) == 0:
+        message = "Runtime environment variable AWS_REGION is not set."
+        logger.critical(message)
+        raise Exception(message)
+    logger.info(f"Got environment variable for AWS_REGION_VAR = {aws_region}")
```
bhazuka, line 50: Update the message should be `AWS_REGION` you may also want to have this be a debug message instead of info.

bhazuka (changes requested):
Overall looks good, A couple of minor issues to fix up. You may want to compare this against the copy in the release branch `release/v3.0.0 as there are some issues that needed to be fixed with the messaging.

tasks/shared_libraries/recovery/shared_recovery.py
rizbi-hassan: changed to  
``` logger.debug(f"Got environment variable for AWS_REGION = {aws_region}")```

tasks/shared_libraries/recovery/README.md
rizbi-hassan: added

tasks/shared_libraries/recovery/shared_recovery.py
rizbi-hassan: removed one

AndrewDorn (changes requested):
We should have unit tests that change based on this, right?

rizbi-hassan:
> We should have unit tests that change based on this, right?

yes you are correct. I will need to patch the "AWS_REGION" env variable in the unit tests orelse it's giving me errors.

AndrewDorn (approved):

bhazuka (approved):

PR #128, opened 2021-08-11
Feature/orca 236

AndrewDorn:
## Summary of Changes

Made `aws_profile` non-required. Will now accept `null`. 

Addresses [ORCA-236: As a developer, I would like to investigate changes needed if the aws_profile variable is not required to better support CI/CD functions.](https://bugs.earthdata.nasa.gov/browse/ORCA-236)

## Changes

* aws_profile will now default to 'default' if not given a value or given a `null` value.

## Type of change

## How Has This Been Tested?

Please describe the tests (unit, integration, ad-hoc, or otherwise) that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Run against AWS. Resources stayed where expected.

## Checklist:

    - [x] CHANGELOG.md
    - [x] User documentation
    - [ ] Unit tests
    - [ ] Integration tests
    - [x] Snyk
    - [x] git-secrets

### DIFF
--- a/modules/orca/main.tf
+++ b/modules/orca/main.tf
@@ -11,7 +11,7 @@ terraform {
 
 ## AWS Provider Settings
 provider "aws" {
-  profile = var.aws_profile
+  profile = local.modified_aws_profile
   region  = var.region
 }
 
@@ -19,6 +19,7 @@ provider "aws" {
 ## Local Variables
 locals {
   tags = merge(var.tags, { Deployment = var.prefix })
+  modified_aws_profile = "${var.aws_profile == null ? "default" : var.aws_profile}"
 }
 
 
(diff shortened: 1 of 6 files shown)


modules/orca/variables.tf
```diff
@@ -4,6 +4,7 @@
 variable "aws_profile" {
   type        = string
   description = "AWS profile used to deploy the terraform application."
+  default = null
```
AndrewDorn, line 7: Could also have the default be `default` but this does communicate clearly to Cumulus that they can use 'null' again. Thoughts?

modules/orca/main.tf
```diff
@@ -11,14 +11,15 @@ terraform {
 
 ## AWS Provider Settings
 provider "aws" {
-  profile = var.aws_profile
```
rizbi-hassan, line 14: do we have to change this `aws_profile` var in the [main.tf](https://github.com/nasa/cumulus-orca/blob/develop/main.tf#L15) as well?

rizbi-hassan (commented):
Just added a comment regarding the variable change. Also suggest running `terraform validate` to check if it returns any error.

modules/orca/main.tf
AndrewDorn: No. This is a local variable.

AndrewDorn:
> Just added a comment regarding the variable change. Also suggest running `terraform validate` to check if it returns any error.

I ran against AWS.

PR #177, opened 2021-11-04
added api gateway for reporting lambda-ORCA 256

rizbi-hassan:
## Summary of Changes

Addresses [ORCA-256: Update API Gateway TF for reconcilliation](https://bugs.earthdata.nasa.gov/browse/ORCA-256)

## Changes

* updated CHANGELOG.md
* updated modules/lambdas/main.tf
* updated website/docs/developer/research/research-APIGateway.md

## Type of change


AndrewDorn:
Should we pull in the API stuff for the status lambdas as well? Would make sense to do it all at once, since the implementation is identical.

rizbi-hassan:
> Should we pull in the API stuff for the status lambdas as well? Would make sense to do it all at once, since the implementation is identical.

yes I can add that as well.

website/docs/developer/deployment-guide/deployment-with-cumulus.md
```diff
@@ -483,8 +483,10 @@ variables is shown in the table below.
 The orca module provides the outputs seen below in the table. Outputs are
 accessed using terraform dot syntax in the format of `module.orca.variable_name`.
 
-| Output Variable                               | Description                               |
-| --------------------------------------------- | ----------------------------------------- |
+| Output Variable                                      | Description                               |
+| -----------------------------------------------------|---------------------------------------------------------|
+| `orca_catalog_reporting_api_invoke_url`              |The URL to invoke the API for catalog reporting lambda |
+| `orca_lambda_copy_to_glacier_cumulus_translator_arn` | AWS ARN of the ORCA orca_lambda_copy_to_glacier_cumulus_translator lambda. |
```
AndrewDorn, line 489: This lambda no longer exists.

AndrewDorn (approved):
One little thing that I'm guessing came from merge-conflicts. Good thought with isolation API gateway stuff at the bottom of the file. Could see making that its own module if it gets bigger.

website/docs/developer/deployment-guide/deployment-with-cumulus.md
rizbi-hassan: removed

modules/lambdas/main.tf
```diff
@@ -479,3 +479,300 @@ data "aws_lambda_invocation" "db_migration" {
...
+  integration_http_method = "POST"
+  type                    = "AWS"
+  uri                     = aws_lambda_function.orca_catalog_reporting.invoke_arn
+}
+
+resource "aws_api_gateway_method_response" "orca_catalog_reporting_response_200" {
+  rest_api_id = aws_api_gateway_rest_api.orca_catalog_reporting_api.id
+  resource_id = aws_api_gateway_resource.orca_catalog_reporting_api_resource.id
+  http_method = aws_api_gateway_method.orca_catalog_reporting_api_method.http_method
+  status_code = "200"
```
bhazuka, line 519: Does this mean we will always return a 200 response even for errors?

modules/lambdas/main.tf
```diff
@@ -479,3 +479,300 @@ data "aws_lambda_invocation" "db_migration" {
...
+}
+
+resource "aws_api_gateway_deployment" "orca_catalog_reporting_api_deployment" {
+  rest_api_id = aws_api_gateway_rest_api.orca_catalog_reporting_api.id
+}
+
+resource "aws_api_gateway_stage" "orca_catalog_reporting_api_stage" {
+  deployment_id = aws_api_gateway_deployment.orca_catalog_reporting_api_deployment.id
+  rest_api_id   = aws_api_gateway_rest_api.orca_catalog_reporting_api.id
+  stage_name    = "dev"
```
bhazuka, line 550: I wonder if this should not be changed "dev" implies development. This should probably be something like orca or orca_production or something like that to denote that it would be a production ready stage.

modules/lambdas/main.tf
```diff
@@ -479,3 +479,300 @@ data "aws_lambda_invocation" "db_migration" {
...
+resource "aws_api_gateway_rest_api" "orca_catalog_reporting_api" {
+  name        = "${var.prefix}_orca_catalog_reporting_api"
+  description = "API for catalog reporting lambda function"
+  endpoint_configuration {
+    types = ["PRIVATE"]
+    # Cumulus might need to create vpc_endpoint_ids which should be added here in the future in order to access this API
+  }
+}
+resource "aws_api_gateway_resource" "orca_catalog_reporting_api_resource" {
+  path_part   = "{proxy+}"
```
bhazuka, line 493: This looks odd to me. shouldn't this be "resource"?

modules/lambdas/output.tf
```diff
@@ -62,4 +62,25 @@ output "post_copy_request_to_queue_arn" {
 output "db_deploy_function_name" {
   description = "AWS Function Name for the db_deploy lambda."
   value       = aws_lambda_function.db_deploy.function_name
+}
+
+# API gateway for catalog reporting lambda
+# ------------------------------------------------------------------------------
+output "catalog_reporting_api_invoke_url" {
```
bhazuka, line 69: See comments from modules/lambdas/main.tf. I am thinking we should just have a single base URL instead of three.

outputs.tf
```diff
@@ -52,6 +52,20 @@ output "orca_lambda_post_copy_request_to_queue_arn" {
   value       = module.orca.orca_lambda_post_copy_request_to_queue_arn
 }
 
+output "orca_catalog_reporting_api_invoke_url" {
```
bhazuka, line 55: See comments from modules/lambdas/main.tf. I am thinking we should just have a single base URL instead of three.

modules/orca/output.tf
```diff
@@ -47,6 +47,22 @@ output "orca_lambda_post_copy_request_to_queue_arn" {
   value       = module.orca_lambdas.post_copy_request_to_queue_arn
 }
 
+output "orca_catalog_reporting_api_invoke_url" {
```
bhazuka, line 50: See comments from modules/lambdas/main.tf. I am thinking we should just have a single base URL instead of three.

website/docs/developer/deployment-guide/deployment-with-cumulus.md
```diff
@@ -483,8 +483,9 @@ variables is shown in the table below.
 The orca module provides the outputs seen below in the table. Outputs are
 accessed using terraform dot syntax in the format of `module.orca.variable_name`.
 
-| Output Variable                               | Description                               |
-| --------------------------------------------- | ----------------------------------------- |
+| Output Variable                                      | Description                               |
+| -----------------------------------------------------|---------------------------------------------------------|
+| `orca_catalog_reporting_api_invoke_url`              |The URL to invoke the API for catalog reporting lambda |
```
bhazuka, line 488: See comments from modules/lambdas/main.tf. I am thinking we should just have a single base URL instead of three.

bhazuka (changes requested):
I think this can be simplified to a single API gateway that connects the three lambdas via pathing. We should include some additional documentation that explains the path, inputs and expected results for the API or at least create a card to perform that work. Similar to security. 

I added a bunch of comments to modules/lambdas/main.tf but they apply to the whole document. I only made the comments once since the rest is copy/paste.

modules/lambdas/main.tf
```diff
@@ -479,3 +479,300 @@ data "aws_lambda_invocation" "db_migration" {
...
+  depends_on  = [aws_api_gateway_integration.orca_catalog_reporting_api_integration]
+  rest_api_id = aws_api_gateway_rest_api.orca_catalog_reporting_api.id
+  resource_id = aws_api_gateway_resource.orca_catalog_reporting_api_resource.id
+  http_method = aws_api_gateway_method.orca_catalog_reporting_api_method.http_method
+  status_code = aws_api_gateway_method_response.orca_catalog_reporting_response_200.status_code
+
+  # Transforms the backend JSON response to XML
+  response_templates = {
+    "application/json" = <<EOF
+    #set($inputRoot = $input.path('$'))
```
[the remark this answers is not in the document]
rizbi-hassan: it was there to transform backend JSON response to XML. I don't think it is being used but kept it since it was in this PR 
https://github.com/nasa/cumulus-orca/pull/102

modules/lambdas/main.tf
```diff
@@ -479,3 +479,300 @@ data "aws_lambda_invocation" "db_migration" {
...
+  rest_api_id = aws_api_gateway_rest_api.orca_catalog_reporting_api.id
+
+  policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "AWS": "*"
```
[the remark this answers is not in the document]
rizbi-hassan: yes that can be done. Do you know the VPC id?

modules/lambdas/main.tf
rizbi-hassan: I can change this to `orca`

modules/lambdas/main.tf
```diff
@@ -479,3 +479,300 @@ data "aws_lambda_invocation" "db_migration" {
 }
 
 ## TODO: Should create null resource to handle password changes ORCA-145
+
+# API Gateway- API for orca_catalog_reporting lambda function
+resource "aws_api_gateway_rest_api" "orca_catalog_reporting_api" {
```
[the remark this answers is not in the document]
rizbi-hassan: I believe we want to have a REST API so used `aws_api_gateway_rest_api`
According to https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/apigatewayv2_api,
`Amazon API Gateway Version 2 resources are used for creating and deploying WebSocket and HTTP APIs. To create and deploy REST APIs, use Amazon API Gateway Version 1`


yes I can move the api stuff to a new folder modules/apigateway/main.tf if that makes it easier to maintain

modules/lambdas/main.tf
rizbi-hassan: I think it will only work if status code is 200 orelse it will return error

modules/lambdas/main.tf
rizbi-hassan: not  sure about this. This shows as {proxy+} according to @AndrewDorn PR https://github.com/nasa/cumulus-orca/pull/102

modules/lambdas/main.tf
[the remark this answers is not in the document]
bhazuka: Note the second part `and HTTP APIs` by using v2 we can do both so gives us some more flexibility in the future.

modules/lambdas/main.tf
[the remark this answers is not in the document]
bhazuka: I would remove it not being used.

modules/lambdas/main.tf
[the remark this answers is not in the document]
bhazuka: It would need to be a variable since the VPC ID would be different depending on who is deploying and to what environment

modules/lambdas/main.tf
bhazuka: Any way to test that and see if for example a 500 error is raised?

modules/lambdas/main.tf
bhazuka: May want to do more investigation. I think proxy is a catchall and maybe not what we want to do here for each one.

modules/lambdas/main.tf
rizbi-hassan: not that I know of at the moment but will take a look at this

modules/lambdas/main.tf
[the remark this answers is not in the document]
rizbi-hassan: after doing some research and testing, `aws_apigatewayv2_api ` does not have some features like selecting endpoint configuration type like `aws_api_gateway_rest_api `. I suggest using the rest api for now for this particular task and look at v2 api later if necessary in a different card.

modules/lambdas/main.tf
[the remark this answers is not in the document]
rizbi-hassan: changed to a required variable in 
https://github.com/nasa/cumulus-orca/pull/177/commits/6ff644eb25669e242fd1b954b5bcd9cb59051ea9

rizbi-hassan:
will update research/research-APIGateway.md once all the PR comments are resolved.

AndrewDorn (changes requested):
Just realized... Why did we move away from the response_templates?

AndrewDorn (approved):

modules/api-gateway/main.tf
```diff
@@ -0,0 +1,227 @@
+# API Gateway- API for ORCA cumulus reconciliation
+resource "aws_api_gateway_rest_api" "orca_cumulus_reconciliation_api" {
```
bhazuka, line 2: Probably want to generalize the api name to `orca_api` instead of `orca_cumulus_reconciliation_api` since we are consolidating to one API.

modules/api-gateway/main.tf
```diff
@@ -0,0 +1,227 @@
...
+      }
+    }
+  ]
+}
+EOF
+}
+
+#API details for orca_catalog_reporting lambda
+resource "aws_api_gateway_resource" "orca_catalog_reporting_api_resource" {
+  path_part   = "orca_catalog_reporting"
```
bhazuka, line 46: Should probably discuss this further. I think we would want a `path` instead of `path_part` and we would want to follow RESTFUL API standards.

modules/api-gateway/main.tf
```diff
@@ -0,0 +1,227 @@
...
+  statement_id  = "AllowExecutionFromAPIGateway"
+  action        = "lambda:InvokeFunction"
+  function_name = "${var.prefix}_orca_catalog_reporting"
+  principal     = "apigateway.amazonaws.com"
+  source_arn    = "${aws_api_gateway_rest_api.orca_cumulus_reconciliation_api.execution_arn}/*/${aws_api_gateway_method.orca_catalog_reporting_api_method.http_method}${aws_api_gateway_resource.orca_catalog_reporting_api_resource.path}"
+}
+
+# API details for  for request_status_for_granule lambda function
+resource "aws_api_gateway_resource" "request_status_for_granule_api_resource" {
+  path_part   = "recovery_status_granules"
```
bhazuka, line 105: Should probably discuss this further. I think we would want a `path` instead of `path_part` and we would want to follow RESTFUL API standards.

modules/api-gateway/main.tf
```diff
@@ -0,0 +1,227 @@
...
+  statement_id  = "AllowExecutionFromAPIGateway"
+  action        = "lambda:InvokeFunction"
+  function_name = "${var.prefix}_request_status_for_granule"
+  principal     = "apigateway.amazonaws.com"
+  source_arn    = "${aws_api_gateway_rest_api.orca_cumulus_reconciliation_api.execution_arn}/*/${aws_api_gateway_method.request_status_for_granule_api_method.http_method}${aws_api_gateway_resource.request_status_for_granule_api_resource.path}"
+}
+
+# API details for request_status_for_job lambda function
+resource "aws_api_gateway_resource" "request_status_for_job_api_resource" {
+  path_part   = "recovery_status_jobs"
```
bhazuka, line 164: Should probably discuss this further. I think we would want a `path` instead of `path_part` and we would want to follow RESTFUL API standards.

bhazuka (changes requested):
I still think we may want to take a closer look at v2 as we probably do not need the extra overhead of V1 plus the v2 API is cheaper and made for lambdas. Overall looks good but a few minor items like variable names and pathing probably need a little discussion.

modules/lambdas/main.tf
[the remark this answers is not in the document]
bhazuka: Not sure that is really needed, plus V2 is cheaper than V1 and made for Lambda. See https://www.serverless.com/blog/api-gateway-v2-http-apis and https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html

modules/lambdas/main.tf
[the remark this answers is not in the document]
bhazuka: Never mind, looking into templates, I see now that is how code logic is added

modules/lambdas/main.tf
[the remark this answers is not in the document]
bhazuka: See notes on the variable and reuse of the Cumulus vpc_id variable instead of creating a new variable.

CHANGELOG.md
```diff
@@ -87,6 +93,7 @@ variable "db_host_endpoint" {
   db_admin_password   = var.db_admin_password
   db_user_password    = var.db_user_password
   db_host_endpoint    = var.db_host_endpoint
+  api_gateway_policy_vpc_id = var.api_gateway_policy_vpc_id
```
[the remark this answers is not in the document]
rizbi-hassan: yes my bad I didn't think that way. Removed api_gateway_policy_vpc_id and used existing vpc_id from everywhere.

modules/lambdas/output.tf
```diff
@@ -56,31 +56,25 @@ output "post_copy_request_to_queue_arn" {
   value       = aws_lambda_function.post_copy_request_to_queue.arn
 }
 
-
-# Utility Lambdas
-# ------------------------------------------------------------------------------
-output "db_deploy_function_name" {
-  description = "AWS Function Name for the db_deploy lambda."
-  value       = aws_lambda_function.db_deploy.function_name
+output "orca_catalog_reporting_invoke_arn" {
```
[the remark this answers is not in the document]
rizbi-hassan: added

modules/api-gateway/main.tf
rizbi-hassan: changed to `orca_api`

modules/lambdas/main.tf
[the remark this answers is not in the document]
rizbi-hassan: from previous conversations with cumulus, I think we want to have a private api and that is only possible using REST api. There is no option to configure as private api when using http v2 api
https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html

modules/lambdas/main.tf
rizbi-hassan: changed the {proxy} to `orca_catalog_reporting`, `recovery_status_granules` and `recovery_status_jobs`

tried using / instead of _ but that is not allowed in the format standards.

rizbi-hassan:
created [orca-295](https://bugs.earthdata.nasa.gov/browse/ORCA-295) for adding detailed api docs.

modules/lambdas/main.tf
rizbi-hassan: after talking with Andrew, you can test this by hitting code that does `return create_http_error_dict`

main.tf
```diff
@@ -33,6 +33,7 @@ module "orca" {
   orca_default_bucket       = var.orca_default_bucket
   
   ## OPTIONAL
+  api_vpc_endpoint_id                                  = var.api_vpc_endpoint_id
```
bhazuka, line 36: May want to call this just `vpc_endpoint_id` to remove confusion that this is the API Gateway id when in reality it is the NGAP VPC endpoint ID.

modules/orca/variables.tf
```diff
@@ -175,4 +175,9 @@ variable "staged_recovery_queue_message_retention_time_seconds" {
 variable "status_update_queue_message_retention_time_seconds" {
   type        = number
   description = "The number of seconds status_update_queue SQS retains a message in seconds. Maximum value is 14 days."
+}
+
+variable "api_vpc_endpoint_id" {
```
bhazuka, line 180: May want to call this just `vpc_endpoint_id` to remove confusion that this is the API Gateway id when in reality it is the NGAP VPC endpoint ID.  See previous notes on description recommendations.

variables.tf
```diff
@@ -194,4 +194,9 @@ variable "status_update_queue_message_retention_time_seconds" {
   type        = number
   description = "The number of seconds status_update_queue SQS retains a message in seconds. Maximum value is 14 days."
   default     = 777600 #9 days
+}
+
+variable "api_vpc_endpoint_id" {
```
bhazuka, line 199: May want to call this just `vpc_endpoint_id` to remove confusion that this is the API Gateway id when in reality it is the NGAP VPC endpoint ID. See previous notes on description updates

website/docs/developer/research/research-APIGateway.md
```diff
@@ -85,13 +85,6 @@ description: Research Notes on API Gateway.
...
     resource "aws_api_gateway_deployment" "[name]_api_deployment" {
       rest_api_id = aws_api_gateway_rest_api.[name]_api.id
-    }
-    ```
-  - aws_api_gateway_stage
-  ```
-    resource "aws_api_gateway_stage" "[name]_api_stage" {
-      deployment_id = aws_api_gateway_deployment.[name]_api_deployment.id
-      rest_api_id   = aws_api_gateway_rest_api.[name]_api.id
       stage_name    = "dev" #add your stage name
```
bhazuka, line 88: may want to change the stage name here

bhazuka (changes requested):
Questions and some thoughts on the variable name and some confusion that may occur.

modules/api-gateway/main.tf
```diff
@@ -42,23 +52,29 @@ EOF
 }
 
 #API details for orca_catalog_reporting lambda
-resource "aws_api_gateway_resource" "orca_catalog_reporting_api_resource" {
-  path_part   = "orca_catalog_reporting"
+resource "aws_api_gateway_resource" "orca_catalog_reporting_api_resource_catalog" {
+  path_part   = "catalog"
```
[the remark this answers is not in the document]
rizbi-hassan: for `aws_api_gateway_resource`, path_part is a required property and there is no attribute named `path`
in `path_part`, I tried using `catalog/reconciliation` but that gave error 
`Error creating API Gateway Resource: BadRequestException: Resource's path part only allow a-zA-Z0-9._-: or a valid greedy path variable and curly braces at the beginning and the end.`

that's why created a different resource.

website/docs/developer/research/research-APIGateway.md
rizbi-hassan: changed to `orca`

variables.tf
rizbi-hassan: changed to vpc_endpoint_id

modules/orca/variables.tf
rizbi-hassan: changed to vpc_endpoint_id

main.tf
rizbi-hassan: changed to vpc_endpoint_id

modules/orca/main.tf
```diff
@@ -141,4 +141,6 @@ module "orca_api_gateway" {
   orca_catalog_reporting_invoke_arn     = module.orca_lambdas.orca_catalog_reporting_invoke_arn
   request_status_for_granule_invoke_arn = module.orca_lambdas.request_status_for_granule_invoke_arn
   request_status_for_job_invoke_arn     = module.orca_lambdas.request_status_for_job_invoke_arn
+  ## OPTIONAL
+  api_vpc_endpoint_id                      = var.api_vpc_endpoint_id
```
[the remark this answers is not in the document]
rizbi-hassan: changed to vpc_endpoint_id

modules/api-gateway/variables.tf
```diff
@@ -29,4 +29,10 @@ variable "api_gateway_stage_name" {
   type        = string
   description = "stage name for the ORCA cumulus reconciliation api gateway"
   default     = "orca"
+}
+
+variable "api_vpc_endpoint_id" {
+  type        = string
+  description = "api gateway vpc endpoint id needed to access the api. Defaults to null."
```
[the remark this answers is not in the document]
rizbi-hassan: changed to vpc_endpoint_id and updated description to 
`NGAP vpc endpoint id needed to access the api. Defaults to null`

bhazuka (approved):

PR #187, opened 2021-11-24
Feature/orca 258

AndrewDorn:
## Summary of Changes

Addresses [ORCA-258: Research Spike-Solidify Architecture for Internal Reconciliation](https://bugs.earthdata.nasa.gov/browse/ORCA-258)

## Changes

* Refined documentation for internal reconciliation.


website/static/files/ads_orca_schema.xed
```diff
@@ -1,616 +1,604 @@
-<?xml version="1.0" encoding="UTF-8"?><Diagram version="1.0" type="POSTGRES" versionMajor="9" versionMinor="6" groupSchema="false" hide_inv="false">
```
AndrewDorn, line 1: Looks like diff doesn't work too well here. I promise I only removed provider_id from reconcile_catalog_mismatch_report.

website/docs/developer/research/research-reconciliation.mdx
```diff
@@ -692,48 +678,71 @@ items for managing jobs.
...
+| s3_last_update        | `timestamptz` | LastModified key value from the S3 listing. | NULLABLE column |
+| orca_size             | `int8`    | Size value from the ORCA File object the file belongs to. | From Catalog. |
+| s3_size               | `int8`    | Size key value from the S3 listing.       | NULLABLE column |
+| discrepancy_type      | `string`  | Type of discrepancy encountered like "missing" or "value x mismatch" | |
 
 
 ## Open Questions yet to be answered
 
 - What is the proper date to use when receiving a date range from Cumulus for a reconciliation report?
+  - Answer: Milliseconds since Unix Epoch.
```
bhazuka, line 745: May want to put that the format is milliseconds since unix epoch. A date range for the reconciliation from Cumulus would use the `createdAt` date from Cumulus that is part of the granule record. The comparison would use the last update internally, I would think at the file level.

bhazuka (approved):
Overall looks good to me. Maybe a few minor tweaks for clarification.

rizbi-hassan (approved):

website/docs/developer/research/research-reconciliation.mdx
AndrewDorn: I thought this was just about the format for some reason. Removed for now.

PR #195, opened 2021-12-08
Feature/orca 259

rizbi-hassan:
## Summary of Changes

Addresses [ORCA-259: Convert Bamboo Plans to Code](https://bugs.earthdata.nasa.gov/browse/ORCA-259)

## Changes

* Created bambo-specs/bamboo.yaml definition file.

## Type of change

## How Has This Been Tested?

tested previously in feature/ORCA-259 branch before changing the default branch to develop. Plan was able to build successfully in bamboo.

## What is the expected behaviour?
Once code is merged into develop, bamboo should automatically detect the new plan file  in`bamboo-specs/bamboo.yaml` and then deploy the pipeline since Webhook is now added.

The old CI plan "ORCA Integrator" will be disabled and a new plan will be created once merged into develop

### DIFF
--- a/bamboo-specs/bamboo.yaml
+++ b/bamboo-specs/bamboo.yaml
@@ -284,7 +284,7 @@ branches:
 ---
 version: 2
 plan:
-  key: ORCA-PP
+  key: ORCA-OI
 plan-permissions:
 - users:
   - bhazuka
--- a/bamboo-specs/bamboo.yaml
+++ b/bamboo-specs/bamboo.yaml
@@ -49,7 +49,6 @@ Build and Test:
     volumes:
       ${bamboo.working.directory}: ${bamboo.working.directory}
       ${bamboo.tmp.directory}: ${bamboo.tmp.directory}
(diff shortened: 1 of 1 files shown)


AndrewDorn (approved):
I see the bits I'd expect.

bhazuka (approved):
I noticed that the script in YAML is sometimes has the commands quoted and sometimes not like in lines 130 and lines 59 for example. Is there a way to make it consistent throughout to match the style on line 130 so it is more readable?

rizbi-hassan:
> I noticed that the script in YAML is sometimes has the commands quoted and sometimes not like in lines 130 and lines 59 for example. Is there a way to make it consistent throughout to match the style on line 130 so it is more readable?

When the script is written in one line using newline `\n`, it needs to have "". 

I can modify the scripts into different lines to use the format as in line 130 if that's desirable. 

Bamboo created the scripts in yaml file in that format by default.

bhazuka (approved):

PR #203, opened 2021-12-10
updated post_to_catalog to match new cumulus schema

rizbi-hassan:
## Summary of Changes

Addresses [ORCA-318: Update post_to_catalog with new catalog schema](https://bugs.earthdata.nasa.gov/browse/ORCA-318)

## Changes

* updated tasks/post_to_catalog/post_to_catalog.py
* updated tasks/post_to_catalog/schemas/catalog_record_input.json
* updated tasks/post_to_catalog/test/unit_tests/test_post_to_catalog.py
* updated CHANGELOG.md

## Type of change

## How Has This Been Tested?

ran unit tests.

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
    - [x] CHANGELOG.md


tasks/post_to_catalog/schemas/catalog_record_input.json
```diff
@@ -63,6 +64,10 @@
         "files"
       ],
       "properties": {
+        "providerId": {
```
AndrewDorn, line 67: This is already present on the provider property

tasks/post_to_catalog/test/unit_tests/test_post_to_catalog.py
```diff
@@ -83,6 +83,7 @@ def test_send_record_to_database_happy_path(
             "version": uuid.uuid4().__str__(),
         }
         granule = {
+            "providerId": uuid.uuid4().__str__(),
```
AndrewDorn, line 86: Per my comment on the schema, these conflict with provider["providerId"]. Tests shouldn't need any additional input.

tasks/post_to_catalog/post_to_catalog.py
```diff
@@ -107,21 +107,12 @@ def create_catalog_records(
...
-                        "collection_id": collection["collectionId"]
-                    }
-                ]
-            )
             LOGGER.debug("Creating granule record '{cumulusGranuleId}'", cumulusGranuleId=granule["cumulusGranuleId"])
             results = connection.execute(
                 create_granule_sql(),
                 [
                     {
+                        "provider_id": provider["providerId"],
```
AndrewDorn, line 115: This is the correct way to get the providerId.

AndrewDorn (changes requested):
You can use the existing providerId property.

tasks/post_to_catalog/schemas/catalog_record_input.json
rizbi-hassan: but from the new cumulus schema, we have provider_id under granules table. So shouldn't this be added as well here?

tasks/post_to_catalog/test/unit_tests/test_post_to_catalog.py
rizbi-hassan: changed to `"provider_id": provider["providerId"]`

tasks/post_to_catalog/post_to_catalog.py
rizbi-hassan: gotcha

tasks/post_to_catalog/post_to_catalog.py
```diff
@@ -197,26 +188,13 @@ def create_collection_sql():
...
         """
     INSERT INTO granules
-        (collection_id, cumulus_granule_id, execution_id, ingest_time, cumulus_create_time, last_update)
+        (provider_id, collection_id, cumulus_granule_id, execution_id, ingest_time, cumulus_create_time, last_update)
     VALUES
-        (:collection_id, :cumulus_granule_id, :execution_id, :ingest_time, :cumulus_create_time, :last_update)
+        (:provider_id, :collection_id, :cumulus_granule_id, :execution_id, :ingest_time, :cumulus_create_time, :last_update)
     ON CONFLICT (collection_id, cumulus_granule_id) DO UPDATE
         SET 
             execution_id=:execution_id, last_update=:last_update
```
bhazuka, line 200: I wonder if we shouldn't also update the provider_id here along with the execution_id. Thoughts? @AndrewDorn  and @rizbi-hassan

tasks/post_to_catalog/schemas/catalog_record_input.json
```diff
@@ -55,6 +55,7 @@
       "type": "object",
       "description": "A dictionary containing granule information",
       "required": [
+        "providerId",
```
bhazuka, line 58: I don not believe we want this here. If we do, we should get rid of the provider block above. as a result, we would also need to change copy_to_gralcier to work with the changes to the SQS schema here. I feel like since we have the provider block above, we are ok.

tasks/post_to_catalog/schemas/catalog_record_input.json
bhazuka: I do not believe so. I would put it in just one place. We may want to discuss changing the provider schema block to being a singular value instead of an array though. If we do, we will also need to make adjustments to copy_to_glacier as well to conform to the change.

tasks/post_to_catalog/test/unit_tests/test_post_to_catalog.py
```diff
@@ -83,6 +83,7 @@ def test_send_record_to_database_happy_path(
             "version": uuid.uuid4().__str__(),
         }
         granule = {
+            "providerId": provider["providerId"],
```
bhazuka, line 86: Depending on what we decide with the input schema, this may need to be removed.

tasks/post_to_catalog/test/unit_tests/test_post_to_catalog.py
```diff
@@ -147,6 +148,7 @@ def test_send_record_to_database_rejects_bad_record(
             "version": uuid.uuid4().__str__(),
         }
         granule = {
+            "providerId": uuid.uuid4().__str__(),
```
bhazuka, line 151: Depending on what we decide with the input schema, this may need to be removed.

tasks/post_to_catalog/test/unit_tests/test_post_to_catalog.py
```diff
@@ -218,6 +218,7 @@ def test_create_catalog_records_happy_path(
             "etag": uuid.uuid4().__str__(),
         }
         granule = {
+            "providerId": uuid.uuid4().__str__(),
```
bhazuka, line 221: Depending on what we decide with the input schema, this may need to be removed.

bhazuka (changes requested):
We should probably gather up as a team and discuss the input schema a little bit as it will affect the remaining code. Lets try to do it after or morning tag up.

tasks/post_to_catalog/schemas/catalog_record_input.json
rizbi-hassan: then I would prefer removing the provider id from here so that we don't need to modify copy_to_glacier

rizbi-hassan: removed

tasks/post_to_catalog/schemas/catalog_record_input.json
rizbi-hassan: removed providerId under granules in the schema

tasks/post_to_catalog/test/unit_tests/test_post_to_catalog.py
rizbi-hassan: removed

tasks/post_to_catalog/test/unit_tests/test_post_to_catalog.py
rizbi-hassan: removed

tasks/post_to_catalog/test/unit_tests/test_post_to_catalog.py
rizbi-hassan: removed

tasks/post_to_catalog/post_to_catalog.py
AndrewDorn: Ah, maybe. Don't know if Cumulus would throw a modified provider_id for a granule at us.

rizbi-hassan: added provider_id under `ON CONFLICT`

AndrewDorn (approved):

tasks/post_to_catalog/post_to_catalog.py
bhazuka: That is not the correct place. The conflict here is based on a unique constraint that states the collection and granule id combination must be unique. Per my comment this would update the provider id value along with the execution id and date.

should be 

```
ON CONFLICT (collection_id, cumulus_granule_id) DO UPDATE
        SET 
            provider_id=:provider_id, execution_id=:execution_id, last_update=:last_update
  RETURNING id"""
```

bhazuka (changes requested):
Need to update the upsert with the correct syntax.

bhazuka (approved):

PR #220, opened 2022-01-07
Feature/orca 161

rizbi-hassan:
## Summary of Changes

Addresses [ORCA-161: Enhance SQS deployments with a DLQ and Cloud watch alarms](https://bugs.earthdata.nasa.gov/browse/ORCA-161)

## Changes

* updated modules/sqs/main.tf
*  updated shared library recovery code to differentiate between standard and fifo queue
* updated changelog.md

## Type of change

## How Has This Been Tested?

tested by deploying in aws. Sent incorrect values to the original recovery queue. Once there were 4 messages, they were sent to the dead letter queue after 1.5 hours due to visibility timeout being set to 30mins. The cloudwatch alarm got triggered and I received an email with alarm details.

## Checklist:

    - [x] CHANGELOG.md


modules/sqs/main.tf
```diff
@@ -53,17 +53,81 @@ resource "aws_sqs_queue" "metadata_queue" {
...
+  message_retention_seconds  = var.staged_recovery_queue_message_retention_time_seconds
+  tags                       = local.tags
+  policy                     = data.aws_iam_policy_document.staged_recovery_queue_policy.json
+  visibility_timeout_seconds = 1800 # Set to double lambda max time
+  depends_on = [
+    aws_sqs_queue.staged_recovery_deadletter_queue
+  ]
+  redrive_policy = jsonencode({
+    deadLetterTargetArn = aws_sqs_queue.staged_recovery_deadletter_queue.arn,
+    maxReceiveCount     = 3 #if number of delayed messages in the queue is greater than this number, then the messages will be sent to its DLQ.
```
rizbi-hassan, line 68: TBD with team on what number to use here. Currently I set it to 3 which means when there are >3 messages stuck in SQS, they will be sent to dead letter queue to trigger the alarm.

modules/sqs/main.tf
```diff
@@ -53,17 +53,81 @@ resource "aws_sqs_queue" "metadata_queue" {
...
+# SNS topic needed for cloudwatch alarm
+resource "aws_sns_topic" "staged_recovery_dlq_alarm_topic" {
+  name = "staged_recovery_deadletter_queue_alarm_topic"
+}
+
+resource "aws_sns_topic_subscription" "sns_subscription_for_alarm" {
+  depends_on = [aws_sns_topic.staged_recovery_dlq_alarm_topic]
+  topic_arn  = aws_sns_topic.staged_recovery_dlq_alarm_topic.arn
+  protocol   = "email"
+  endpoint   = "rhassan@contractor.usgs.gov"
```
rizbi-hassan, line 129: I am guessing it will be an “email” protocol. If so, TBD on who to add to the email.

modules/sqs/main.tf
AndrewDorn: According to docs, https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.htm this is actually the number of times a message can be pulled from the queue and rejects (in our case by a lambda error) before it is sent to the dead letter queue.

modules/sqs/main.tf
```diff
@@ -53,17 +53,81 @@ resource "aws_sqs_queue" "metadata_queue" {
...
+resource "aws_cloudwatch_metric_alarm" "staged_recovery_deadletter_alarm" {
+  alarm_name          = "${aws_sqs_queue.staged_recovery_deadletter_queue.name}-not-empty-alarm"
+  alarm_description   = "Items are on the ${aws_sqs_queue.staged_recovery_deadletter_queue.name} queue"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods  = 1
+  metric_name         = "ApproximateNumberOfMessagesVisible"
+  namespace           = "AWS/SQS"
+  period              = 300 #In seconds. Valid values for period are 1, 5, 10, 30, or any multiple of 60. 
+  statistic           = "Average"
+  threshold           = 1 #alarm will be triggered if number of messages in the DLQ equals to this threshold.
```
AndrewDorn, line 111: Not strictly accurate. The statistic is "Average", not sum. So if a single message is in the queue and does not remain for the entire period, the threshold will not be tripped.

modules/sqs/main.tf
AndrewDorn: Could see an argument that these should come directly to us. Could also see an argument that this should be configurable, so the user can get a notification that something they tried failed. If the former, we should use an email address that targets our group. If the later, should add configuration.

modules/sqs/main.tf
```diff
@@ -53,17 +53,81 @@ resource "aws_sqs_queue" "metadata_queue" {
...
+    ]
+  }
+}
+
+resource "aws_cloudwatch_metric_alarm" "staged_recovery_deadletter_alarm" {
+  alarm_name          = "${aws_sqs_queue.staged_recovery_deadletter_queue.name}-not-empty-alarm"
+  alarm_description   = "Items are on the ${aws_sqs_queue.staged_recovery_deadletter_queue.name} queue"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods  = 1
+  metric_name         = "ApproximateNumberOfMessagesVisible"
```
AndrewDorn, line 107: Might vote for NumberOfMessagesReceived, with a comparison_operator of sum. That way messages stop being received if the queue stops filling up. As-is, emails will be sent out once per period for as long as the queue remains non-empty.

modules/sqs/main.tf
```diff
@@ -53,17 +53,81 @@ resource "aws_sqs_queue" "metadata_queue" {
...
-  visibility_timeout_seconds  = 1800 # Set to double lambda max time
+  name                       = "${var.prefix}-staged-recovery-queue"
+  delay_seconds              = var.sqs_delay_time_seconds
+  max_message_size           = var.sqs_maximum_message_size
+  message_retention_seconds  = var.staged_recovery_queue_message_retention_time_seconds
+  tags                       = local.tags
+  policy                     = data.aws_iam_policy_document.staged_recovery_queue_policy.json
+  visibility_timeout_seconds = 1800 # Set to double lambda max time
+  depends_on = [
+    aws_sqs_queue.staged_recovery_deadletter_queue
```
AndrewDorn, line 64: Should be able to remove this due to the deadletter_queue being used/depended on in the redrive_policy.

shared_libraries/orca_shared/recovery/shared_recovery.py
```diff
@@ -191,18 +191,30 @@ def post_entry_to_queue(
     md5_body = hashlib.md5(body.encode("utf8")).hexdigest()  # nosec
 
     LOGGER.debug("Sending message to the QUEUE")
-    response = mysqs.send_message(
-        QueueUrl=db_queue_url,
-        MessageDeduplicationId=deduplication_id,
```
AndrewDorn, line 196: Just to be clear, we're fine with losing deduplication? Does something else take care of that?

shared_libraries/orca_shared/recovery/shared_recovery.py
```diff
@@ -191,18 +191,30 @@ def post_entry_to_queue(
...
-        MessageGroupId="request_files",
-        MessageAttributes={
-            "RequestMethod": {
-                "DataType": "String",
-                "StringValue": request_method.value,
-            }
-        },
-        MessageBody=body,
-    )
+    if ".fifo" in db_queue_url:
```
AndrewDorn, line 194: Really not a fan of code in the shared_recovery being used for queues beyond the recovery queue. This type of branching should not be necessary. If we really want posting to the status queue and posting to the copy queue in the same lib, then we can separate this into two functions and use the fifo-only function without a RequestMethod for post_copy_request_to_queue.

AndrewDorn (changes requested):
Couple concerns with how we handle error poling, and bloat on post_entry_to_queue.

shared_libraries/orca_shared/recovery/shared_recovery.py
rizbi-hassan: yes we can create a new function for fifo queue but that means adding duplicate code in the shared lib. that's why decided to use `if else `statement

shared_libraries/orca_shared/recovery/shared_recovery.py
rizbi-hassan: deduplication is only required for FIFO queue. since this now a standard queue, that is not needed.

modules/sqs/main.tf
rizbi-hassan: I got circular dependency without adding this. The DLQ has to be created first before the original queue.

modules/sqs/main.tf
rizbi-hassan: yes `NumberOfMessagesReceived` should also work. That could be an alternative.

From my testing, everything looked good with the above config. I only get one alarm if there are message in DLQ.
In your suggested config, I am assuming we will continue getting alarm emails every time period which might fill our inbox.

modules/sqs/main.tf
rizbi-hassan: I was talking about `maxReceiveCount`. See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sqs_queue#redrive_policy

modules/sqs/main.tf
rizbi-hassan: it will send an alarm when threshold is 1. So actually average or sum will give same result.
see example
https://github.com/damacus/terraform-aws-sqs-with-dlq/blob/a041c4343d99a3c46a53dce8ac0fda2bc6559193/main.tf#L87

modules/sqs/main.tf
AndrewDorn: Yes. The page you linked links to the page I linked under the description for redrive_policy. From that aws doc: "The maxReceiveCount is the number of times a consumer tries receiving a message from a queue without deleting it before being moved to the dead-letter queue."

modules/sqs/main.tf
AndrewDorn: I don't know what this link is intended to suggest. My understanding is that if one item is in the queue for 299 seconds, then the Average would be 1 * (299/300) which would not trigger the threshold.

modules/sqs/main.tf
AndrewDorn: Is there something built into the alarm that stops emails from coming in? As written:
If the queue has more than one element, trigger alarm.
What I'm suggesting:
If the queue receives more than one element, trigger alarm.
The former will trigger for as long as the queue remains non-empty, the later will trigger only when new items are added to the queue. Unless there's something where this alarm can only trigger emails once until the queue is empty again.

shared_libraries/orca_shared/recovery/shared_recovery.py
AndrewDorn: As-is deduplication_id, and request_method both live in limbo, required without actually being required for all cases. I think the fact that request_method is an unused parameter is enough to split this into two functions. You can pull 178-183 and 218-231 into shared functions, then do

```
mysqs = setup_sqs_connection(db_queue_url)

md5_body = hashlib.md5(body.encode("utf8")).hexdigest()  # nosec
LOGGER.debug("Sending message to the QUEUE")
response = mysqs.send_message(unique code isolated)

check_sqs_response_code(response_
```

AndrewDorn: We will want eventually to completely separate non-shared code for posting to the copy queue, as it is not fully related to the status queue.

modules/sqs/main.tf
rizbi-hassan: aah good catch. fixed the description

modules/sqs/main.tf
rizbi-hassan: makes sense. I will have to test more to see what happens with sum.

If everyone agrees, I can change this to `sum`

modules/sqs/main.tf
rizbi-hassan: right now, if the number of messages in DLQ is `GreaterThanOrEqualToThreshold ` which is 1, it will send an alarm only once.
when the queue is empty again and then it receives a message, it will again send an alarm

shared_libraries/orca_shared/recovery/shared_recovery.py
rizbi-hassan: ok makes sense. I can separate the functions. anything you want to add @bhazuka ?

modules/sqs/main.tf
AndrewDorn: Ah, gotcha. Thanks for the explanation.

modules/sqs/main.tf
bhazuka: I think sum may be a better choice here.

modules/sqs/main.tf
bhazuka: In the short term, I would make the endpoint email a variable so end users can add whatever email/group email they want. We will want to bring this up with the working group and discuss how they want to handle DLQ items and expectations. We may want to even consider other items like can we retry the DLQ messages (retryable error) and if the message cannot be retried, update our database with the error information for job status and delete the message as possible future work.

shared_libraries/orca_shared/recovery/shared_recovery.py
bhazuka: Looking over the code, this looks to be a base function. I agree it will be easier to maintain to break into two separate base functions, one for FIF queues and one for standard. This would also allow you to remove other items not needed like `MessageAttributes` I could also see where this could be split into three functions if you really wanted.
- A base SQS function that is generic that creates the client, gets the md5hash of the body and sends the message and validates using **kwargs so that it could be used generically anywhere. This is a function that could be used in a more generic utility library later.
- A function for Fifo that creates the dedupe info and adds Message Attributes, Group ID, etc then calls the base function 
- A Function for standard that just sends the QueueUrl and body

modules/sqs/main.tf
rizbi-hassan: so this will be a new required orca variable, right? @bhazuka

rizbi-hassan: added a new required orca variable for the email

shared_libraries/orca_shared/recovery/shared_recovery.py
rizbi-hassan: added a new function for the standard queue. used the existing function for the fifo queue

CHANGELOG.md
```diff
@@ -15,6 +15,99 @@ and includes an additional section for migration notes.
...
+### Added
+- *ORCA-161* Added dead letter queue and cloudwatch alarm terraform code to recovery SQS queue.
+
+### Changed
+- *ORCA-161* Changed staged recovery SQS queue type from FIFO to standard queue.
+
+### Migration Notes
+
+- The user should update their `orca.tf`, `variables.tf` and `terraform.tfvars` files with new variable. The following required variable has been added:
+  - db_admin_username (defaults to "postgres")
```
bhazuka, line 29: Shouldn't this be `dead_letter_queue_topic_subscription_email`? As a side note, @AndrewDorn  and @rizbi-hassan how do you feel about shortening the name of the variable to something like `dlq_subscription_email`?

modules/orca/main.tf
```diff
@@ -134,14 +134,18 @@ module "orca_sqs" {
   ## Cumulus Variables
   ## --------------------------
   ## REQUIRED
-  prefix = var.prefix
+  prefix                                     = var.prefix
```
AndrewDorn, line 137: Unneeded whitespace.

CHANGELOG.md
```diff
@@ -15,6 +15,99 @@ and includes an additional section for migration notes.
...
+
+### Added
+- *ORCA-161* Added dead letter queue and cloudwatch alarm terraform code to recovery SQS queue.
+
+### Changed
+- *ORCA-161* Changed staged recovery SQS queue type from FIFO to standard queue.
+
+### Migration Notes
+
+- The user should update their `orca.tf`, `variables.tf` and `terraform.tfvars` files with new variable. The following required variable has been added:
```
AndrewDorn, line 28: Copy/paste/merge issue? db_admin_username addition exists below.

CHANGELOG.md
```diff
@@ -15,6 +15,99 @@ and includes an additional section for migration notes.
...
+  - db_admin_username (defaults to "postgres")
+  
+- Add the following ORCA required variables definition to your `variables.tf` or `orca_variables.tf` file.
+
+```terraform
```
AndrewDorn, line 33: Same duplicate-doc issue with a lot of this.

main.tf
```diff
@@ -34,6 +34,7 @@ module "orca" {
   db_host_endpoint          = var.db_host_endpoint
   db_user_password          = var.db_user_password
   orca_default_bucket       = var.orca_default_bucket
+  dead_letter_queue_topic_subscription_email = var.dead_letter_queue_topic_subscription_email
```
AndrewDorn, line 37: Should probably fix the alignment above this.

modules/sqs/main.tf
```diff
@@ -53,17 +53,81 @@ resource "aws_sqs_queue" "metadata_queue" {
...
+  ]
+  redrive_policy = jsonencode({
+    deadLetterTargetArn = aws_sqs_queue.staged_recovery_deadletter_queue.arn,
+    maxReceiveCount     = 3 #number of times a consumer tries receiving a message from the queue without deleting it before being moved to DLQ.
+  })
 }
 
+resource "aws_sqs_queue" "staged_recovery_deadletter_queue" {
```
bhazuka, line 72: may think of shortening from `staged_recovery_deadletter_queue` to `staged_recovery_dlq`

modules/sqs/main.tf
```diff
@@ -53,17 +53,81 @@ resource "aws_sqs_queue" "metadata_queue" {
...
+    "QueueName" = aws_sqs_queue.staged_recovery_deadletter_queue.name
+  }
+}
+
+# SNS topic needed for cloudwatch alarm
+resource "aws_sns_topic" "staged_recovery_dlq_alarm_topic" {
+  name = "staged_recovery_deadletter_queue_alarm_topic"
+}
+
+resource "aws_sns_topic_subscription" "sns_subscription_for_alarm" {
```
bhazuka, line 125: may want to change name from `sns_subscription_for_alarm` to `dlq_alarm_email`

CHANGELOG.md
```diff
@@ -15,6 +15,99 @@ and includes an additional section for migration notes.
...
+}
+
+variable "rds_security_group_id" {
+  type        = string
+  description = "Cumulus' RDS Security Group's ID."
+}
+
+variable "dead_letter_queue_topic_subscription_email" {
+  type        = string
+  description = "The email to notify users when messages are received in dead letter SQS queue"
```
AndrewDorn, line 56: Should just need this part. Might want to emphasize that the queue is filled by failed restore requests, and that emails will only happen once until the dead letter queue is emptied. Might be good to note that in general docs as well.

modules/sqs/main.tf
```diff
@@ -53,17 +53,81 @@ resource "aws_sqs_queue" "metadata_queue" {
...
+  treat_missing_data  = "notBreaching"
+  alarm_actions       = [aws_sns_topic.staged_recovery_dlq_alarm_topic.arn]
+  tags                = local.tags
+  dimensions = {
+    "QueueName" = aws_sqs_queue.staged_recovery_deadletter_queue.name
+  }
+}
+
+# SNS topic needed for cloudwatch alarm
+resource "aws_sns_topic" "staged_recovery_dlq_alarm_topic" {
```
bhazuka, line 121: May want to change name from `staged_recovery_dlq_alarm_topic` ro `staged_recovery_dlq_alarm` see Nates wiki notes https://wiki.earthdata.nasa.gov/display/ORCA/ORCA+Terraform+Overview+and+Best+Practices on best practices.

shared_libraries/orca_shared/recovery/__init__.py
```diff
@@ -5,6 +5,6 @@
     RequestMethod,
     create_status_for_job,
     get_aws_region,
-    post_entry_to_queue,
+    post_entry_to_fifo_queue,
```
bhazuka, line 8: Need to add `post_entry_to_standard_queue` to the list.

modules/sqs/main.tf
```diff
@@ -53,17 +53,81 @@ resource "aws_sqs_queue" "metadata_queue" {
...
+  name = "staged_recovery_deadletter_queue_alarm_topic"
+}
+
+resource "aws_sns_topic_subscription" "sns_subscription_for_alarm" {
+  depends_on = [aws_sns_topic.staged_recovery_dlq_alarm_topic]
+  topic_arn  = aws_sns_topic.staged_recovery_dlq_alarm_topic.arn
+  protocol   = "email"
+  endpoint   = var.dead_letter_queue_topic_subscription_email   #todo: ORCA-365
```
AndrewDorn, line 129: Not a problem, let's just remember to remove this todo if no changes arise from discussions with Cumulus.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -171,7 +166,6 @@ def exponential_delay(base_delay: int, exponential_backoff: int = 2) -> int:
 def query_db(key_path: str, bucket_name: str) -> List[Dict[str, str]]:
     """
     Connect and query the recover_file status table return needed metadata for posting to the recovery status SQS Queue.
-
```
bhazuka, line 174: Not sure why you are removing these empty lines. They provide whitespace making the description easier to read.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -239,10 +232,8 @@ def query_db(key_path: str, bucket_name: str) -> List[Dict[str, str]]:
 def get_metadata_sql(key_path: str) -> text:
     """
     Query for finding metadata based on key_path and PENDING status.
-
```
bhazuka, line 242: Not sure why you are removing these empty lines. They provide whitespace making the description easier to read.

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
```diff
@@ -264,14 +255,12 @@ def handler(event: Dict[str, Any], context) -> None:
     """
     Lambda handler. This lambda calls the task function to perform db queries
     and send message to SQS.
-
```
bhazuka, line 267: Not sure why you are removing these empty lines. They provide whitespace making the description easier to read.

shared_libraries/orca_shared/recovery/shared_recovery.py
```diff
@@ -217,3 +217,50 @@ def post_entry_to_queue(
...
+    mysqs = mysqs_resource.Queue(recovery_queue_url)
+
+    md5_body = hashlib.md5(body.encode("utf8")).hexdigest()  # nosec
+
+    LOGGER.debug("Sending message to the QUEUE")
+    response = mysqs.send_message(
+        QueueUrl=recovery_queue_url,
+        MessageBody=body,
+    )
+    LOGGER.debug("SQS Message Response: {response}", response=json.dumps(response))
```
AndrewDorn, line 253: Everything from here down could be put in a shared `check_sqs_response` method, but I'm also okay with keeping things fully separate, at least for now.

shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
```diff
@@ -55,18 +55,18 @@ def tearDown(self):
...
         function returns the same expected message.
         """
         new_data = {"name": "test"}
         for request_method in self.request_methods:
             # Run subtests
             with self.subTest(request_method=request_method):
                 # Send values to the function
-                shared_recovery.post_entry_to_queue(
+                shared_recovery.post_entry_to_fifo_queue(
                     new_data, request_method, self.db_queue_url
```
AndrewDorn, line 70: We might want a card on the backlog to add to these tests. Hopefully there's a way to check that request_method is used.

bhazuka (changes requested):
Overall looking good, couple of items missing and some thoughts on styling and naming.

tasks/post_copy_request_to_queue/test/unit_tests/test_post_copy_request_to_queue.py
```diff
@@ -342,7 +340,7 @@ def test_task_happy_path(
                 post_copy_request_to_queue.TARGET_KEY_KEY: key_path,
                 post_copy_request_to_queue.SOURCE_BUCKET_KEY: bucket_name,
             }
-        ]
+        mock_query_db.return_value = [row]
```
AndrewDorn, line 343: We have cases where lambdas modify input before using it, so that's why I did this excessive multi-initiation. This way is better so long as we do`mock_query_db.return_value = [copy.deepcopy(row)]
See line 410 for an example.

AndrewDorn (commented):
Just doc stuff now.

CHANGELOG.md
rizbi-hassan: yes my bad.. removed and changed variable name to dlq_subscription_email

modules/sqs/main.tf
rizbi-hassan: updated.

modules/sqs/main.tf
rizbi-hassan: changed

modules/sqs/main.tf
rizbi-hassan: changed

shared_libraries/orca_shared/recovery/__init__.py
rizbi-hassan: oops added

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: added back

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: added back

tasks/post_copy_request_to_queue/post_copy_request_to_queue.py
rizbi-hassan: added back

modules/orca/main.tf
rizbi-hassan: removed

CHANGELOG.md
rizbi-hassan: oops fixed

CHANGELOG.md
rizbi-hassan: removed duplicates

main.tf
rizbi-hassan: fixed

CHANGELOG.md
rizbi-hassan: ok I thought we needed to add all the required variables here. 
removed the other variables.

Added more info to the variable description like you suggested

modules/sqs/main.tf
rizbi-hassan: yes sir

tasks/post_copy_request_to_queue/test/unit_tests/test_post_copy_request_to_queue.py
rizbi-hassan: ok used mock_query_db.return_value = [copy.deepcopy(row)]

shared_libraries/orca_shared/recovery/shared_recovery.py
rizbi-hassan: sounds good.

shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
rizbi-hassan: i think request_method is captured in MessageAttributes and then checked

                queue_contents = self.queue.receive_messages(
                    MessageAttributeNames=["All"]
                )
                queue_output_body = json.loads(queue_contents[0].body)

                # Testing SQS body
                self.assertEqual(queue_output_body, new_data)

AndrewDorn: Nope. Note how new_data is just `{"name": "test"}`
We'll want to check the error cases eventually as well.

rizbi-hassan: sounds good. created [ORCA-366](https://bugs.earthdata.nasa.gov/browse/ORCA-366) for additional tests.

AndrewDorn (approved):

modules/orca/main.tf
```diff
@@ -144,7 +144,7 @@ module "orca_sqs" {
   ## --------------------------
 
   ## REQUIRED
-  dead_letter_queue_topic_subscription_email           = var.dead_letter_queue_topic_subscription_email
+  dlq_subscription_email                               = var.dlq_subscription_email
```
bhazuka, line 147: Some weird spacing going on here.

bhazuka (approved):

modules/orca/main.tf
rizbi-hassan: removed

PR #227, opened 2022-01-13
Removed deleted file from build_release_package

AndrewDorn:


bhazuka (approved):

website/docs/developer/development-guide/code/versioning-releases.md
```diff
@@ -49,18 +49,19 @@ bamboo deployment plan.
 1. The deployment plan is already created in Bamboo using [Bamboo Specs](https://github.com/nasa/cumulus-orca/tree/develop/bamboo-specs) 
    In the ORCA project (https://ci.earthdata.nasa.gov/browse/ORCA-OI), scroll to the top left of the page where it indicates `Plan branch`. From the `Plan branch` dropdown menu, select the release branch you created for the release which should be in the format `release-X.X.X`.
 2. Once inside the release branch page, scroll to the top right of the page and click `Actions`-> `Configure branch`.
-3. On the `Plan branch configuration` page, click on the `Variables` tab. Ensure that you are on your branch plan and not the
-   master plan. Click on the `Choose from inherited variables` dropdown menu. Set a DEPLOYMENT variable appropriate for the
-   release (defaults to last committer). This should be cumulus-from-npm-tf 
+3. On the `Plan branch configuration` page, under `Plan branch configuration`, enable 'Change Trigger'. Set the 
+   Trigger type to manual, and this will prevent commits to the branch from triggering the build plan.
```
rizbi-hassan, line 53: I thought we wanted to use "repository polling" here as trigger type. This was the default config even before changing to the bamboo specs.

Any changes committed to the branch will trigger the job and run the pipeline to see if the unit tests are working fine. Isn't that what we want?

rizbi-hassan (changes requested):

website/docs/developer/development-guide/code/versioning-releases.md
rizbi-hassan: right now the trigger type is manual for the `Release` stage though.

yes it makes sense that doing what you said will not run the jobs every time a change is commited and pushed to develop.

But we have to manually run all stages of the pipeline every time instead of just the "Release" stage in that case.

rizbi-hassan: my assumption was that we want to automatically run all stages upto Release and then run that Release stage as manual which is what we have currently. 
@bhazuka any thoughts on this?

PR #228, opened 2022-01-14
Add deployment fixes

markdboyd:


modules/lambdas/main.tf
```diff
@@ -456,6 +461,8 @@ resource "aws_lambda_function" "db_deploy" {
 data "aws_lambda_invocation" "db_migration" {
   depends_on    = [aws_lambda_function.db_deploy]
   function_name = aws_lambda_function.db_deploy.function_name
-  input         = jsonencode({})
+  input         = jsonencode({
+    replacementTrigger = timestamp()
+  })
```
markdboyd, line 466: This should prevent the lambda invocation from running during the `terraform plan` phase and only run it during the `terraform apply` phase. This is important because otherwise if you have somehow misconfigured your deployment, subsequent deployments will keep running this Lambda invocation with your old configuration before your new configuration can be deployed, creating an endless loop that you can't get out of.

modules/orca/main.tf
```diff
@@ -154,6 +154,9 @@ module "orca_sqs" {
 ## orca_api_gateway - api gateway module
 ## =============================================================================
 module "orca_api_gateway" {
+  depends_on = [
+    module.orca_lambdas
+  ]
```
markdboyd, line 159: This is necessary because the API gateway module depends on a Lambda function existing:

https://github.com/nasa/cumulus-orca/blob/release-4.0.0-Beta3/modules/api-gateway/main.tf#L238-L244

But without this change, there is nothing in the code that will force the API gateway module to wait to deploy until the lambdas module is deployed, which was causing errors for me.

AndrewDorn (approved):

rizbi-hassan (approved):

PR #259, opened 2022-03-30
Feature/orca 373 Create Inventory Report bucket and allow proper access.

AndrewDorn:
## Summary of Changes

Addresses [ORCA-373: Create Inventory Report bucket and allow proper access.](https://bugs.earthdata.nasa.gov/browse/ORCA-373)

## Changes

* Added report bucket to TF.

## Type of change

## How Has This Been Tested?

Deployed and run in AWS.

## Checklist:

    - [x] CHANGELOG.md
    - [x] Testing documentation
    - [ ] User documentation
    - [x] Snyk
    - [x] git-secrets

### DIFF
--- a/modules/orca/main.tf
+++ b/modules/orca/main.tf
@@ -11,6 +11,7 @@ module "orca_s3" {
   buckets                                              = var.buckets
   prefix                                               = var.prefix
   ## OPTIONAL
+  s3_report_frequency                                  = var.s3_report_frequency
   tags                                                 = var.tags
 }
 
--- a/modules/orca/main.tf
+++ b/modules/orca/main.tf
@@ -1,10 +1,18 @@
-## Local Variables
-locals {
-  tags = merge(var.tags, { Deployment = var.prefix })
-}
-
-
 ## Referenced Modules
+
+## s3 - Automated bucket/inventory report creation
+## =============================
+module "orca_s3" {
+  source = "../s3"
+  ## --------------------------
+  ## Cumulus Variables
+  ## --------------------------
+  ## REQUIRED
+  buckets                                              = var.buckets
+  prefix                                               = var.prefix
+  ## OPTIONAL
+  tags                                                 = var.tags
+}
 
 ## orca_lambdas - lambdas module that calls iam and security_groups module
 ## =============================
@@ -15,13 +23,13 @@ module "orca_lambdas" {
   ## Cumulus Variables
   ## --------------------------
   ## REQUIRED
-  buckets                                              = var.buckets
+  buckets                                              = module.orca_s3.buckets
   lambda_subnet_ids                                    = var.lambda_subnet_ids
   prefix                                               = var.prefix
   rds_security_group_id                                = var.rds_security_group_id
   vpc_id                                               = var.vpc_id
   ## OPTIONAL
-  tags                           = local.tags
+  tags                           = var.tags
   default_multipart_chunksize_mb = var.default_multipart_chunksize_mb
 
   ## --------------------------
@@ -67,7 +75,7 @@ module "orca_lambdas_secondary" {
   lambda_subnet_ids                                    = var.lambda_subnet_ids
   prefix                                               = var.prefix
   ## OPTIONAL
-  tags                           = local.tags
+  tags                           = var.tags
   default_multipart_chunksize_mb = var.default_multipart_chunksize_mb
 
   ## --------------------------
@@ -99,7 +107,7 @@ module "orca_workflows" {
   workflow_config = var.workflow_config
 
   ## OPTIONAL
-  tags = local.tags
+  tags = var.tags
 
   ## --------------------------
   ## ORCA Variables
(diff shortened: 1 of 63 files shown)


AndrewDorn:
I was unable to locate documentation on creating buckets.

tasks/db_deploy/install/orca_sql.py
```diff
@@ -234,10 +234,6 @@ def create_extension() -> TextClause:
         """
             -- Create extension
             CREATE EXTENSION IF NOT EXISTS aws_s3 CASCADE;
-
-            -- Comment
-            COMMENT ON EXTENSION aws_s3
-              IS 'Custom AWS extension that allows for data COPY from and to an S3 bucket object.';
```
AndrewDorn, line 240: Did this so I could get things working. We can hold off if desired.

modules/orca/variables.tf
```diff
@@ -89,6 +89,12 @@ variable "orca_default_bucket" {
 }
 
 
+variable "orca_reports_bucket" {
```
rizbih, line 92: modules/orca/main.tf also needs to be updated

AndrewDorn: It will eventually, but presently nothing consumes this value. We'll see what comes out of the automated report generation.

AndrewDorn:
> how are we actually creating the bucket? I was expecting something like adding a new module like `modules/s3/main.tf` that will have the terraform code to deploy the bucket.
> 
> Also bucket policy is missing.

Currently all of our buckets are created prior to deployment, possibly to make sure that contents are not accidentally wiped during redeploy/adjustments. Definitely open to changing that if desired.

AndrewDorn:
> how are we actually creating the bucket? I was expecting something like adding a new module like `modules/s3/main.tf` that will have the terraform code to deploy the bucket.
> 
> Also bucket policy is missing.

I am unable to locate

AndrewDorn:
Looks like I missed some requirements. Closing for now.

modules/orca/variables.tf
AndrewDorn: Removed pending further discussion.

modules/s3/main.tf
```diff
@@ -0,0 +1,40 @@
+## Local Variables
+locals {
+  orca_bucket_names  = [for k, v in var.buckets : v.name if v.type == "orca"]
+}
+
+resource "aws_s3_bucket" "reports_bucket" {
```
rizbih, line 6: is there any requirement to enable versioning in this bucket?

rizbih, line 6: I think we can also add an `aws_s3_bucket_policy` resource that will be attached to this bucket if we want to make it more restricted

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_policy

modules/s3/main.tf
```diff
@@ -0,0 +1,40 @@
...
+  for_each = toset(local.orca_bucket_names)
+
+  bucket = each.key
+  name   = "${var.prefix}-${each.key}-inventory-report"
+
+  included_object_versions = "All"
+  optional_fields = ["Size", "LastModifiedDate", "StorageClass", "ETag"]
+
+  schedule {
+    frequency = "Daily"
```
rizbih, line 30: If we want to save cost, we can also consider generating the reports weekly if acceptable.

modules/orca/main.tf
```diff
@@ -122,11 +130,11 @@ module "orca_iam" {
   ## Cumulus Variables
   ## --------------------------
   ## REQUIRED
-  buckets                  = var.buckets
+  buckets                  = module.orca_s3.buckets
```
rizbih, line 133: I think this should also be updated 
https://github.com/nasa/cumulus-orca/blob/4d6cff91690984c4a8b947d89a2115d1e0009301/main.tf#L17

maybe verify with `terraform validate`

rizbih (changes requested):
some comments on terraform regarding creating the bucket

modules/s3/main.tf
AndrewDorn: Nah. Different versions of the same report exist on different keys.

modules/s3/main.tf
AndrewDorn: Good call. Pulled out into a variable.

modules/s3/main.tf
AndrewDorn: So you want something where even requests within the same account are denied unless they are on a whitelist? Feels like a bit of overkill for a non-public-facing resource.

rizbih: looking at our other non-public-facing buckets in sandbox, looks like they don't have a bucket policy in them either. So I guess we don't need a policy right now.

versions.tf
```diff
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = ">= 3.63.0"
+      version = ">= 3.63.0, < 4.0.0"
```
AndrewDorn, line 7: 4.0.0 split stuff out of existing resources such as splitting the retention policy out of s3 buckets. If we ever want to upgrade, this will require a task.

PR #261, opened 2022-03-31
Feature/orca 334

clappn:
## Summary of Changes

Addresses [ORCA-334](https://bugs.earthdata.nasa.gov/browse/ORCA-334)

## Changes

* Created new aws_iam_policy_document, aws_iam_role, and aws_iam_role_policy objects used for lambda execution
* Created an aws_iam_role_policy_attachment used for lambda deployment and attached it to the created aws_iam_role
* Set the role to be used by the extract_filepaths_for_granule lambda

## Type of change

- [ ] This change requires a documentation update

## How Has This Been Tested?

- Test the lambda using the lambda execution UI
- Test the lambda using the OrcaRecoveryWorkflow stepfunction

## Checklist:

    - [x] CHANGELOG.md
    - [x] Snyk
    - [x] git-secrets

### DIFF
--- a/modules/lambdas/main.tf
+++ b/modules/lambdas/main.tf
@@ -1,5 +1,6 @@
 # Local Variables
 locals {
+  tags         = merge(var.tags, { Deployment = var.prefix })
   orca_buckets = [for k, v in var.buckets : v.name if v.type == "orca"]
 }
 
@@ -127,7 +128,7 @@ resource "aws_lambda_function" "perform_orca_reconcile" {
 resource "aws_lambda_function" "extract_filepaths_for_granule" {
   ## REQUIRED
   function_name = "${var.prefix}_extract_filepaths_for_granule"
-  role          = var.restore_object_role_arn
+  role		= aws_iam_role.extract_filepaths_for_granule_iam_role.arn
 
   ## OPTIONAL
   description      = "Extracts bucket info and granules filepath from the CMA for ORCA request_files lambda."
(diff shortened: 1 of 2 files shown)


modules/lambdas/main.tf
```diff
@@ -1,5 +1,6 @@
 # Local Variables
 locals {
+  tags         = merge(var.tags, { Deployment = var.prefix })
```
rizbih, line 3: I think you will need to merge develop branch into this branch to pull in the latest changes. 

In the develop branch, it no longer has the local tags value.

modules/lambdas/main.tf
```diff
@@ -146,6 +147,55 @@ resource "aws_lambda_function" "extract_filepaths_for_granule" {
 }
 
 
+data "aws_iam_policy_document" "assume_lambda_role_extract" {
+  statement {
+    principals {
+      type = "AWS"
+      identifiers = ["*"]
```
rizbih, line 154: The identifier should be `["lambda.amazonaws.com"]`

See similar example here https://github.com/nasa/cumulus-orca/blob/4d6cff91690984c4a8b947d89a2115d1e0009301/modules/iam/main.tf#L17

modules/lambdas/main.tf
```diff
@@ -146,6 +147,55 @@ resource "aws_lambda_function" "extract_filepaths_for_granule" {
 }
 
 
+data "aws_iam_policy_document" "assume_lambda_role_extract" {
+  statement {
+    principals {
+      type = "AWS"
```
rizbih, line 153: Should the type be `AWS `or `Service`?
In similar example in orca, it uses Service
https://github.com/nasa/cumulus-orca/blob/4d6cff91690984c4a8b947d89a2115d1e0009301/modules/iam/main.tf#L16

modules/lambdas/main.tf
```diff
@@ -146,6 +147,55 @@ resource "aws_lambda_function" "extract_filepaths_for_granule" {
...
+
+  }
+}
+
+# Permissions needed to run the function successfully
+data "aws_iam_policy_document" "extract_filepaths_for_granule_policy_document" {
+  statement {
+    actions = [
+      "lambda:Invoke*",
+      "s3:PutObject*"
```
rizbih, line 166: are we adding anything to a bucket using that lambda? If not, I think this can be removed.

rizbih (changes requested):
A few things to look into.

modules/lambdas/main.tf
clappn: I changed this back. The original intention was to keep tags self-contained so we wouldn't need to rely on a .tfvars file but it shouldn't matter either way

modules/lambdas/main.tf
```diff
@@ -146,6 +147,55 @@ resource "aws_lambda_function" "extract_filepaths_for_granule" {
...
+
+resource "aws_iam_role_policy" "extract_filepaths_for_granule_policy" {
+  name   = "${var.prefix}_extract_filepaths_for_granule_policy"
+  role   = aws_iam_role.extract_filepaths_for_granule_iam_role.id
+  policy = data.aws_iam_policy_document.extract_filepaths_for_granule_policy_document.json
+}
+
+
+# Attaches permissions needed to deploy the Lambda
+resource "aws_iam_role_policy_attachment" "extract_filepaths_for_granule_iam_policy_attachment" {
```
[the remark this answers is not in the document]
clappn: Removed this and moved the actions that the attachment would give to the `extract_filepaths_for_granule_policy_document` policy document

modules/lambdas/main.tf
clappn: Fixed this, the AWS tech said I needed to have both "lambda.amazonaws.com" and "states.amazonaws.com" set here

modules/lambdas/main.tf
clappn: The lambda does not work at all if "s3:PutObject" isn't defined, I tried using "s3:GetObject" but the function and step function immediately fail.

modules/lambdas/main.tf
rizbih: looks good now

modules/lambdas/main.tf
rizbih: sounds good

modules/lambdas/main.tf
rizbih: okay then that makes sense to add it

rizbih (approved):
looks good to me now. Hope you did some testing by manually triggering the lambda from AWS console to see if it gives any weird permission error.

clappn:
Yep, tested using manual triggers of the lambda and step function, and several deployments/destroys to ensure the lambda and dependencies would be created and run without error

PR #268, opened 2022-04-20
Feature/orca 304 Research using ECS instances instead of lambdas for get_current_archive_list and perform_orca_reconcile.

AndrewDorn:
## Summary of Changes

Research using ECS instances instead of lambdas for get_current_archive_list and perform_orca_reconcile. Added recommendations and sample implementation to docs.

Addresses [ORCA-304: Research using ECS instances instead of lambdas for get_current_archive_list and perform_orca_reconcile.](https://bugs.earthdata.nasa.gov/browse/ORCA-304)

## Changes

* Added researched implementation for running Python via Fargate to docs.
* Added suggestions on code changes for merging get_current_archive_list and perform_orca_reconcile to docs.

## Type of change

## How Has This Been Tested?

Deployed sample code to AWS with the instructions given.

## Checklist:

    - [x] Snyk
    - [x] git-secrets

### DIFF
--- a/website/docs/developer/research/research-lambda-container.md
+++ b/website/docs/developer/research/research-lambda-container.md
@@ -62,7 +62,7 @@ Currently NGAP only allows developers to create a `private` ECR repository which
 
 Details on creating the prototype are shown below:
 
-1. Create an ECR repository from AWS CLI if needed as shown. Check [here](https://docs.aws.amazon.com/AmazonECR/latest/userguide/getting-started-cli.html#cli-create-repository) for additional details on this.
+1. Create an [ECR](https://us-west-2.console.aws.amazon.com/ecr/repositories) repository from AWS CLI if needed as shown. Check [here](https://docs.aws.amazon.com/AmazonECR/latest/userguide/getting-started-cli.html#cli-create-repository) for additional details on this.
 
 ```bash
 aws ecr create-repository \
@@ -90,7 +90,7 @@ COPY requirements.txt  .
 RUN  pip3 install -r requirements.txt --target "${LAMBDA_TASK_ROOT}"
 
 # Set the CMD to your handler (could also be done as a parameter override outside of the Dockerfile)
-CMD [ "test.hambda_handler" ]
+CMD [ "test.lambda_handler" ]
 
 ```
 
@@ -162,7 +162,26 @@ The Orca Internal Reconciliation workflow lambdas require an alternative approac
     - Remember that in addition to processing time, Aurora Serverless can take up to 5 minutes to spin up.
   - Raise the internal-report queue's `visibility_timeout_seconds` to the expected timeout.
   - Environment variables can be passed in at task definition, or when the task is run. The former should be sufficient.
-  
+
+- Use an alternative Dockerfile
+  ```yaml
+  FROM python:3.8-slim-buster
+
+  WORKDIR /src
+
+  # Copy function code
+  COPY main.py main.py
+  COPY sqs_library.py sqs_library.py
+
+  # Install the function's dependencies using file requirements.txt
+  # from your project folder.
+
+  COPY requirements.txt  .
+  RUN  pip3 install -r requirements.txt
+  COPY . .
+
+  CMD ["python", "main.py"]
+  ```
 - Follow prior instructions up to the Terraform deployment.
 - Create an IAM role with the permissions needed. Use case should be `EC2`. Note that developers do not have this capability.
 - Create a task definition for the image.
@@ -170,15 +189,15 @@ The Orca Internal Reconciliation workflow lambdas require an alternative approac
   - `PREFIX-orca-internal-reconciliation-task` name
   - `PREFIX-ecs-task-role` Task role
   - `Linux` Operating system family
-  - Use the previously created IAM role for Task execution IAM role
+  - Task Execution Role must be able to pull container images and publish container logs.
+  - Use the previously created IAM role for Task role
   - Smallest options for Task memory and Task CPU
   - Add container
     - `PREFIX-orca-internal-reconciliation-container` name
     - Link to the uploaded image for Image
 - Create an ECS cluster
-
-TODO: Demo this with a simple service that posts to a queue.
-
+  - Documentation on creation suggests that VPC is required, but developers are not authorized to create.
+- The task can now be run in the cluster. Current tests halt at this point, as the available role cannot pull secrets.
 
 
 
--- a/website/docs/developer/research/research-lambda-container.md
+++ b/website/docs/developer/research/research-lambda-container.md
@@ -137,6 +137,8 @@ Using the steps above, a prototype has been created in NGAP AWS sandbox account 
 ## Future directions
 Currently, the only option to store the Docker image for lambda containers is AWS ECR repository. Github package could be an option to store the image but in order to deploy the lambda, the image has to be stored into an ECR.
 
+The above example Dockerfile does not follow best practices. See example in section on Fargate.
+
 A few possible discussion items:
 - Discuss with NGAP team if the docker images can be stored in a public ECR repo.
 - If github packages are supported to store and deploy the lambda, then we have to contact NASA github admins to enable this feature in our repository.
@@ -165,22 +167,46 @@ The Orca Internal Reconciliation workflow lambdas require an alternative approac
 
 - Use an alternative Dockerfile. The example below packages two code files into a lightweight python container with a `CMD` to run the main file.
   ```yaml
-  FROM python:3.8-slim-buster
+  # Build Stage - Here we can bring dev libraries and compile the wheels for the python libs
+  # =============================================================================
+  FROM python:3.8-slim as builder
 
-  WORKDIR /src
+  WORKDIR /app
+
+  ENV PYTHONDONTWRITEBYTECODE 1
+  ENV PYTHONUNBUFFERED 1
+
+  # Install the function's dependencies using file requirements.txt
+  # from your project folder.
+  COPY requirements.txt  .
+  RUN pip wheel --no-cache-dir --no-deps --wheel-dir /app/wheels -r requirements.txt
+
+  # Run Stage - Contains everything needed to run code with entry point
+  # ===================================================================
+  FROM python:3.8-slim
+
+  # CREATE Non-Privileged User
+  RUN mkdir -p /app \
+      && groupadd -r appuser \
+      && useradd -r -s /bin/false -g appuser appuser \
+      && chown -R appuser:appuser /app
+
+  # Copy compiled wheels and install the requirements
+  COPY --from=builder /app/wheels /wheels
+  RUN pip install --no-cache /wheels/*
+
+  # Set the work directory for our application
+  WORKDIR /app
 
   # Copy function code
   COPY main.py main.py
   COPY sqs_library.py sqs_library.py
 
-  # Install the function's dependencies using file requirements.txt
-  # from your project folder.
+  # Set the User that the app will run as
+  USER appuser
 
-  COPY requirements.txt  .
-  RUN  pip3 install -r requirements.txt
-  COPY . .
-
-  CMD ["python", "main.py"]
+  # Set the entry point for the container
+  ENTRYPOINT ["python", "main.py"]
   ```
 - Follow prior instructions up to the Terraform deployment. Use the following Terraform instead of the previous example.
 ```terraform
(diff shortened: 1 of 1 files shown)


website/docs/developer/research/research-lambda-container.md
```diff
@@ -148,6 +148,152 @@ Based on this research, it looks like using lambda as containers is possible whe
...
+
+### Modifying get_current_archive_list and perform_orca_reconcile
+
+The Orca Internal Reconciliation workflow lambdas require an alternative approach.
+- The maximum time limit for lambdas is 15 minutes. These lambdas may take a significant amount of time, and should not be subject to this limitation.
+
+- Code changes
+  - Merge `get_current_archive_list` and `perform_orca_reconcile` into one codebase.
+  - Wrap functionality in a loop that will process the internal-report queue until no entries remain.
+  - Since ECS does not support timeout, create an overarching timing mechanic that exits if an infinite loop occurs while processing a queue entry.
```
rizbihassan, line 160: could be a problem that it does not support timeout.

website/docs/developer/research/research-lambda-container.md
```diff
@@ -148,6 +148,152 @@ Based on this research, it looks like using lambda as containers is possible whe
...
+  # Install the function's dependencies using file requirements.txt
+  # from your project folder.
+
+  COPY requirements.txt  .
+  RUN  pip3 install -r requirements.txt
+  COPY . .
+
+  CMD ["python", "main.py"]
+  ```
+- Follow prior instructions up to the Terraform deployment. Use the following Terraform instead of the pervious example.
```
rizbihassan, line 185: "previous" spelling.

I use Code Spell checker extension in vscode which is helpful in detecting wrong spellings.

website/docs/developer/research/research-lambda-container.md
```diff
@@ -148,6 +148,152 @@ Based on this research, it looks like using lambda as containers is possible whe
...
+The above example was developed for deploying a simple task that posts to an sqs queue. Names and values should be changed to match new use cases.
+:::
+:::warning
+Applying admin permissions to orca_ecs_task_execution_role is likely overly permissive.
+:::
+- The task can now be run in the cluster.
+This can be done through [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Client.run_task) or the GUI,
+though the former is presently untested.
+
+
```
rizbihassan, line 295: Will it be possible to some info on cost analysis of using ECS instead of lambdas?

I know we "pay as we go" for lambdas which is cheap. But I am not so familiar with ECS cost. Does the cluster run always and if so, how much do we pay per month? 

I think the cost could be a concern.

website/docs/developer/research/research-lambda-container.md
```diff
@@ -148,6 +148,152 @@ Based on this research, it looks like using lambda as containers is possible whe
...
+
 ##### References
```
rizbihassan, line 297: adding some more references on setting up ECS cluster would help.

website/docs/developer/research/research-lambda-container.md
```diff
@@ -148,6 +148,152 @@ Based on this research, it looks like using lambda as containers is possible whe
...
+    capacity_provider = "FARGATE"
+    weight            = 100
+  }
+}
+
+resource "aws_ecs_task_definition" "task" {
+  family                   = "${var.prefix}_orca_sqs_task"
+  network_mode             = "awsvpc"
+  requires_compatibilities = ["FARGATE"]
+  cpu                      = "4096"
```
rizbihassan, line 254: using larger cpu will add in cost. We need to decide a proper value in case fargate is used

rizbihassan (changes requested):
Added some suggestions on adding some more info that could help differentiate ECS vs lambda.

website/docs/developer/research/research-lambda-container.md
```diff
@@ -148,6 +148,152 @@ Based on this research, it looks like using lambda as containers is possible whe
...
+  tags                 = var.tags
+}
+
+
+resource "aws_iam_role_policy_attachment" "ecs_role_policy" {
+  role       = aws_iam_role.orca_ecs_task_execution_role.name
+  policy_arn = "arn:aws:iam::aws:policy/AdministratorAccess"
+}
+
+resource "aws_ecs_cluster" "test-cluster" {
```
rizbihassan, line 241: using fargate would mean we will have to maintain this cluster as well which is a downside compared to using lambdas.

website/docs/developer/research/research-lambda-container.md
AndrewDorn: Fixed.

website/docs/developer/research/research-lambda-container.md
AndrewDorn: So long as we stick with Fargate tasks, we can make use of auto-scaling. Added link.

website/docs/developer/research/research-lambda-container.md
AndrewDorn: I opted to stick with TF so we can build this into our automation flow. See `resource "aws_ecs_cluster" "test-cluster"`.

website/docs/developer/research/research-lambda-container.md
AndrewDorn: Added note.

website/docs/developer/research/research-lambda-container.md
AndrewDorn: True. Unfortunately, lambdas are not an option for the task, and this is the simplest way I found to run a Python script without a time limit.

website/docs/developer/research/research-lambda-container.md
AndrewDorn: Yeah, that one seems weird to me. Not sure why they don't do that.

website/docs/developer/research/research-lambda-container.md
bhazuka: That is actually not true, an ECS Fargate cluster is managed by amazon and automatically scales and does not requires us to update or maintain the underlying OS or structure. If we were to use an ECS EC2 cluster, some additional administration would be required though it is fairly light for our use case.

website/docs/developer/research/research-lambda-container.md
bhazuka: So, I feel like you are looking at this through the view of lambda and not ECS/Docker. In this particular case, this is not really an issue. In the case of the queue, once the queue is empty, the task would stop. In the case of a long running item, other timeouts to cause a failure or force the task to stop would be possible like the database connection timing out. So no real need for an "overarching timing mechanic".

website/docs/developer/research/research-lambda-container.md
rizbihassan: sounds good.

What about the fargate cluster? Do you have to pay for that even when we are not using it?

website/docs/developer/research/research-lambda-container.md
AndrewDorn: Our code within the container can have a bug that causes an infinite loop.

website/docs/developer/research/research-lambda-container.md
AndrewDorn: Fargate refers to a task run under a ECS cluster under the Fargate protocol, so the "fargate container" is just the ECS container. I'll add a couple clarifying details to the doc.

website/docs/developer/research/research-lambda-container.md
AndrewDorn: Technically speaking there are additional components in our TF to maintain, but since they are automated, I'm personally okay with it. At least once we have the repo stuff solidified.

website/docs/developer/research/research-lambda-container.md
```diff
@@ -148,6 +148,156 @@ Based on this research, it looks like using lambda as containers is possible whe
...
+  - Wrap functionality in a loop that will process the internal-report queue until no entries remain.
+  - Since ECS does not support timeout, create an overarching timing mechanic that exits if an infinite loop occurs while processing a queue entry.
+    - Alternatively, a side-program could manually stop the task if it exceeds its' time limit.
+    - Remember that in addition to processing time, Aurora Serverless can take up to 5 minutes to spin up.
+  - Raise the internal-report queue's `visibility_timeout_seconds` to the expected timeout.
+  - Environment variables can be passed in at task definition, or when the task is run. The former should be sufficient.
+
+- Use an alternative Dockerfile. The example below packages two code files into a lightweight python container with a `CMD` to run the main file.
+  ```yaml
+  FROM python:3.8-slim-buster
```
bhazuka, line 168: There are some best practices and some hardening you could put into here for your Dockerfile. Consider the following example code below where we leverage an intermediate build image, create and run as a non root user, and put items that do not change frequently at the top of the build, and those that change more frequently at the bottom. Changes like those proposed lead to a better security (running as non-root so fewer permissions needed) and smaller images (fewer layers and smaller footprint since dev libraries removed). 

```
# Build Stage - Here we can bring dev libraries and compile the wheels for the python libs
# =============================================================================
FROM python:3.8-slim as builder

WORKDIR /app

ENV PYTHONDONTWRITEBYTECODE 1
ENV PYTHONUNBUFFERED 1

# Install compiler and libraries needed for database drivers
RUN apt-get update && \
    apt-get install -y --no-install-recommends gcc libpq-dev libc6-dev

# Install python app requirements
COPY requirements.txt .
RUN pip wheel --no-cache-dir --no-deps --wheel-dir /app/wheels -r requirements.txt


# Final Stage -- This will be the final image
# =============================================================================
FROM python:3.8-slim

# Set metadata information for the image
LABEL application.name="ORCA Internal Reconciliationl" \
    application.description="Reconciles ORCA holdings with what is actually in the S3 Glacier Archive" \
    application.team="ORCA" \
    application.version="v0.0.1"

# CREATE Non-Privileged User
RUN mkdir -p /app \
    && groupadd -r appuser \
    && useradd -r -s /bin/false -g appuser appuser \
    && chown -R appuser:appuser /app

# Install libpq C library for postgres
RUN apt-get update && \
    apt-get install -y --no-install-recommends libpq5

# Copy compiled wheels and install the requirements
COPY --from=builder /app/wheels /wheels
RUN pip install --no-cache /wheels/*

# Set the work directory for our application
WORKDIR /app

# Copy our application source files
COPY *.py .

# Set the User that the app will run as
USER appuser

# Set the entry point for the container
ENTRYPOINT ["python", "main.py"]
```

website/docs/developer/research/research-lambda-container.md
bhazuka: If that is true, it should be discovered via unit testing and/or validation testing and can be shutdown external to the task so it can be fixed.

website/docs/developer/research/research-lambda-container.md
```diff
@@ -148,6 +148,156 @@ Based on this research, it looks like using lambda as containers is possible whe
...
+  requires_compatibilities = ["FARGATE"]
+  cpu                      = "4096"
+  memory                   = "8192"
+  task_role_arn            = aws_iam_role.orca_ecs_tasks_role.arn
+  execution_role_arn       = aws_iam_role.orca_ecs_task_execution_role.arn
+  container_definitions    = <<DEFINITION
+[
+  {
+    "name": "sqs_task_container",
+    "image": "236859827343.dkr.ecr.us-west-2.amazonaws.com/adorn-test-repo:latest",
```
bhazuka, line 263: remove `236859827343` and replace with `999999999999` and instructions to replace with account number throughout.

website/docs/developer/research/research-lambda-container.md
```diff
@@ -148,6 +148,156 @@ Based on this research, it looks like using lambda as containers is possible whe
...
+Applying admin permissions to orca_ecs_task_execution_role is likely overly permissive.
+:::
+- The Fargate task can now be run in the ECS cluster.
+This can be done through [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ecs.html#ECS.Client.run_task) or the GUI,
+though the former is presently untested.
+- [Pricing](https://aws.amazon.com/ecs/pricing/?trk=2f064982-4fad-4e1f-ab75-e1df26258a60&sc_channel=ps&sc_campaign=acquisition&sc_medium=GC-PMM|PS-GO|Brand|All|PA|Database|ECS|Product|US|EN|Text|xx|SEM|PMO22-13405&s_kwcid=AL!4422!3!547620651289!e!!g!!amazon%20ecs%20pricing&ef_id=EAIaIQobChMIsO-ehu2l9wIVCfrICh0gOQ5OEAAYASABEgIZmfD_BwE:G:s&s_kwcid=AL!4422!3!547620651289!e!!g!!amazon%20ecs%20pricing) only applies to what is used in the moment, and can auto-scale down.
+  - Minimum compute time is one minute, so this architecture should not be used for small, frequent tasks.
+
+
+
```
bhazuka, line 300: I would put recommendations in regards to the change (should we move forward or not) and link to any Jira cards to make the necessary changes. That should satisfy the acceptance criteria.

bhazuka (changes requested):
A few thoughts and minor changes. Overall looks really good to me.

website/docs/developer/research/research-lambda-container.md
AndrewDorn: Replaced.

website/docs/developer/research/research-lambda-container.md
AndrewDorn: Added

website/docs/developer/research/research-lambda-container.md
AndrewDorn: Removed

website/docs/developer/research/research-lambda-container.md
AndrewDorn: Added and created.

bhazuka (approved):
Looks good

rizbihassan (approved):

PR #278, opened 2022-05-12
Feature/orca 388 Create Docker image for running ORCA integration tests

rizbihassan:
## Summary of Changes

Addresses [ORCA-388: Create Docker image for running ORCA integration tests ](https://bugs.earthdata.nasa.gov/browse/ORCA-388)

## Changes

* added base_images/Dockerfile

## Type of change

-
## How Has This Been Tested?

- From the base_images/ dir, built the images using
`docker build -f Dockerfile.bamboo  --no-cache --tag bamboo-image .`
`docker build -f Dockerfile.local  --no-cache --tag local-image .`

- Ran the container using
`docker run -it bamboo-image /bin/bash`
`docker run -it local-image /bin/bash`

- Inside the container, verified that the packages are installed

## Checklist:

### DIFF
--- a/base_images/Dockerfile
+++ b/base_images/Dockerfile
@@ -4,7 +4,14 @@ FROM amazonlinux:2
 # TODO: remove packages from tasks/lambda/requirements-dev.txt and tasks/lambda/requirements.txt 
 # except coverage, pytest, fastjsonschema and shared libraries- throws no module found error
 
+ENV NODE_VERSION="16.x"
 ENV TERRAFORM_VERSION "0.13.6"
+
+# Add NodeJS and Yarn repos & update package index
+RUN \
+curl -sL https://rpm.nodesource.com/setup_${NODE_VERSION} | bash - && \
+curl -sL https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo && \
+yum update -y
 
 # CLI utilities
 RUN yum install -y gcc git make openssl wget zip unzip tar
@@ -12,6 +19,7 @@ RUN yum install -y gcc git make openssl wget zip unzip tar
 # install Python 3 and pip
 RUN \
 yum install -y python3-devel && \
+yum install -y nodejs yarn && \
 yum -y install python3-pip && \
 pip3 install -q --upgrade pip --trusted-host pypi.org --trusted-host files.pythonhosted.org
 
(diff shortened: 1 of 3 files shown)


base_images/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+curl "https://s3.amazonaws.com/session-manager-downloads/plugin/latest/linux_64bit/session-manager-plugin.rpm" -o "session-manager-plugin.rpm" && \
+yum install -y session-manager-plugin.rpm
+
+# Install all other necessary packages from lambda requirements-dev.txt and requirements.txt file
+RUN \
+pip3 install psycopg2-binary==2.8.6 SQLAlchemy==1.4.11 pydoc-markdown==4.5.0 boto3==1.18.40 && \
+pip3 install fastjsonschema==2.15.1 cumulus-message-adapter-python==2.0.1 moto[sqs,s3]==2.2.8 && \
+pip3 install bandit==1.7.0 safety==1.10.3 flake8==4.0.1 black==21.12b0 isort==5.10.1
+
+WORKDIR /data
```
rizbihassan, line 55: not sure if this is required.

base_images/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+
+# AWS & Terraform
+RUN \
+yum install -y awscli && \
+wget "https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip" && \
+unzip *.zip && \
+chmod +x terraform && \
+mv terraform /usr/local/bin
+
+# Install and configure git-secrets
```
bhazuka, line 34: We should be able to remove git-secrets from this build since it will be used out in Bamboo land. Git Secretes is good to have when you are doing things locally.

base_images/Dockerfile
bhazuka: Nope, in this context, it should not be needed.

base_images/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+yum install -y session-manager-plugin.rpm
+
+# Install all other necessary packages from lambda requirements-dev.txt and requirements.txt file
+RUN \
+pip3 install psycopg2-binary==2.8.6 SQLAlchemy==1.4.11 pydoc-markdown==4.5.0 boto3==1.18.40 && \
+pip3 install fastjsonschema==2.15.1 cumulus-message-adapter-python==2.0.1 moto[sqs,s3]==2.2.8 && \
+pip3 install bandit==1.7.0 safety==1.10.3 flake8==4.0.1 black==21.12b0 isort==5.10.1
+
+WORKDIR /data
+CMD ["/bin/bash"]
```
bhazuka, line 56: Remove CMD line as it is not needed in this context.

base_images/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+git secrets --install ~/.git-templates/git-secrets && \
+git config --global init.templateDir ~/.git-templates/git-secrets
+
+# SSM SessionManager plugin
+RUN \
+curl "https://s3.amazonaws.com/session-manager-downloads/plugin/latest/linux_64bit/session-manager-plugin.rpm" -o "session-manager-plugin.rpm" && \
+yum install -y session-manager-plugin.rpm
+
+# Install all other necessary packages from lambda requirements-dev.txt and requirements.txt file
+RUN \
```
bhazuka, line 50: I would remove this section. At the most I would only install the test suite stuff
- bandit
- safety
- flake8
- black
- isort
- pydoc-markdown

Everything else is variable and not really needed. I would also think about combining with the pip3 update command as a single layer.

base_images/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+cd /root && \
+git clone https://github.com/awslabs/git-secrets.git && \
+cd git-secrets && \
+make install && \
+git secrets --register-aws --global && \
+git secrets --install ~/.git-templates/git-secrets && \
+git config --global init.templateDir ~/.git-templates/git-secrets
+
+# SSM SessionManager plugin
```
bhazuka, line 44: Consider moving this up and possibly splitting the commands So that you are downloading the RPM/repos in one shot and installing them with yum to minimize the number of layers

bhazuka (changes requested):
There are a couple additional ways we can optimize this image. We should also add some metadata about it like name, maintainer, and version. I wonder if we shouldn't make 2 Dockerfiles. 1 for Bamboo that we use for our pipelines and 1 for local development which would have some of the niceties of this one like git-secrets, the working directory, etc.
Maybe `Dockerfile.bamboo` and `Dockerfile.local`? Something we can talk through tomorrow after standup.

base_images/Dockerfile
rizbihassan: removed

base_images/Dockerfile
rizbihassan: removed

base_images/Dockerfile
rizbihassan: removed except the above packages

AndrewDorn (approved):

bhazuka (approved):
Initial files look good to me.

Issue #285, opened 2022-05-27
psycopg2 error: missing role during database creation


sflynn-itsc:
I was asked to post this issue here by a colleague who will be following up with a pull request. Actual user/role has been "REDACTED" for github: @amarouane-ABDELHAK  here is the issue you requested.

The following error occurs during automated Cumulus deployments (e.g., using Earthdata Bamboo CICD) that include ORCA:
```
[...]

PR #286, opened 2022-05-31
Fixing the issue for Postgres to read camelCase admin username

amarouane-ABDELHAK:
## Summary of Changes

Addresses [issue 285](https://github.com/nasa/cumulus-orca/issues/285)

## Changes

* Grant a role or use admin role to create a database can now use an admin name with camelCase

## Type of change

## How Has This Been Tested?

- [ ] Tested when deployed in GHRC environments  SIT/UAT


AndrewDorn:
If we want this as a patched version, we will have to target the most recent release branch so we don't include the Unreleased changes mentioned in the CHANGELOG.md

PR #297, opened 2022-06-27
Feature/orca 476

rizbihassan:
## Summary of Changes

Addresses [ORCA-476: Deploy ORCA buckets in disaster recovery account.](https://bugs.earthdata.nasa.gov/browse/ORCA-476)

## Changes

* updated bamboo-specs/bamboo.yaml
* created integration_test/deploy-dr-buckets.sh
* created integration_test/dr-buckets.tf.template
* updated docs.

## Type of change

- [ x] This change requires a documentation update

## How Has This Been Tested?

deployed via terraform and bamboo in DR sandbox account. Cross account access verified by accessing DR buckets from cumulus sandbox.

## Checklist:

    - [x ] Development documentation

- [x ] My changes generate no new warnings

### DIFF
--- a/integration_test/deploy-dr-buckets.sh
+++ b/integration_test/deploy-dr-buckets.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+set -ex
+
+export AWS_ACCESS_KEY_ID=$bamboo_AWS_ACCESS_KEY_ID
+export AWS_SECRET_ACCESS_KEY=$bamboo_AWS_SECRET_ACCESS_KEY
+export AWS_DEFAULT_REGION=$bamboo_AWS_DEFAULT_REGION
+
+#remove old files from bamboo as they throw error
+rm *.tf
+git clone --branch develop --single-branch https://github.com/nasa/cumulus-orca.git && cd integration_test
+#replace prefix with bamboo prefix variable
+sed -e 's/PREFIX/'"$bamboo_PREFIX"'/g' dr-buckets.tf.template > dr-buckets.tf
+
+if ! aws s3 cp s3://$bamboo_PREFIX-dr-tf-state/dr-buckets-tf/terraform.tfstate .;then
+    echo "terraform state file not present in S3. Creating the buckets in DR account."
+else
+    echo "Terraform state file found in S3 bucket. Using S3 remote backend"
+fi
+
+terraform init -input=false
+echo "Deploying S3  buckets in Disaaster Recovery account"
+terraform apply \
+  -auto-approve \
+  -input=false
+
+#copy terraform state file to the created tf-state bucket
+aws s3 cp terraform.tfstate s3://$bamboo_PREFIX-dr-tf-state/dr-buckets-tf/terraform.tfstate
--- a/integration_test/dr-buckets.tf.template
+++ b/integration_test/dr-buckets.tf.template
@@ -1,0 +1,97 @@
+resource "aws_s3_bucket" "dr-tf-state" {
+  bucket = "PREFIX-dr-tf-state"
+  force_destroy = true
+  tags = {
+    Deployment = "PREFIX"
+    Application = "ORCA"
+  }
+}
+resource "aws_s3_bucket" "orca-archive-worm" {
+  bucket = "PREFIX-orca-archive-worm"
+  force_destroy = true
+  tags = {
+    Deployment = "PREFIX"
+    Application = "ORCA"
+  }
+}
+resource "aws_s3_bucket_versioning" "orca-archive-worm-versioning" {
+  bucket = aws_s3_bucket.orca-archive-worm.id
+  versioning_configuration {
+    status = "Enabled"
+  }
+}
+resource "aws_s3_bucket" "orca-reports" {
+  bucket = "PREFIX-orca-reports"
+  depends_on = [aws_s3_bucket.orca-archive-worm]
+  force_destroy = true
+  tags = {
+    Deployment = "PREFIX"
+    Application = "ORCA"
+  }
+}
+resource "aws_s3_bucket_lifecycle_configuration" "orca-reports-lifecycle-rule" {
+  bucket = aws_s3_bucket.orca-reports.id
+  rule {
+    id = "delete objects after 30 days"
+    status = "Enabled"
+    expiration {
+      days = 30
+    }
+  }
+}
+# data "aws_iam_policy_document" "orca-reports-policy-json" {
+#   statement {
+#     principals {
+#       type        = "AWS"
+#       identifiers = ["arn:aws:iam::236859827343:root"]
+#     }
+#     actions = [
+#         "s3:GetObject*",
+#         "s3:GetBucket*",
+#         "s3:ListBucket",
+#         "s3:PutObject",
+#         "s3:PutObjectAcl",
+#         "s3:PutBucketNotification"
+#     ]
+#     resources = ["*"]
+#   }
+#   statement {
+#     principals {
+#       type        = "Service"
+#       identifiers = ["s3.amazonaws.com"]
+#     }
+#     actions = [
+#         "s3:PutObject"
+#     ]
+#     resources = ["arn:aws:s3:::orca-archive-worm/*"]
+#   }
+# }
+
+# resource "aws_s3_bucket_policy" "orca-reports-policy" {
+#   bucket = aws_s3_bucket.orca-reports.id
+#   policy = data.aws_iam_policy_document.orca-reports-policy-json.json
+# }
+resource "aws_s3_bucket" "orca-primary" {
+  bucket = "PREFIX-orca-primary"
+  force_destroy = true
+  tags = {
+    Deployment = "PREFIX"
+    Application = "ORCA"
+  }
+}
+# data "aws_iam_policy_document" "orca-primary-policy-json" {
+#  statement {
+#    principals {
+#      type        = "AWS"
+ #     identifiers = ["arn:aws:iam::236859827343:root"]
+#     }
+#     actions = [
+#         "s3:*"
+#     ]
+#     resources = ["arn:aws:s3:::*"]
+#   }
+# }
+# resource "aws_s3_bucket_policy" "orca-primary-policy" {
+#   bucket = aws_s3_bucket.orca-primary.id
+#   policy = data.aws_iam_policy_document.orca-primary-policy-json.json
+# }
--- a/integration_test/dr-buckets.tf.template
+++ b/integration_test/dr-buckets.tf.template
@@ -39,38 +39,95 @@ resource "aws_s3_bucket_lifecycle_configuration" "orca-reports-lifecycle-rule" {
     }
   }
 }
-# data "aws_iam_policy_document" "orca-reports-policy-json" {
-#   statement {
-#     principals {
-#       type        = "AWS"
-#       identifiers = ["arn:aws:iam::236859827343:root"]
-#     }
-#     actions = [
-#         "s3:GetObject*",
-#         "s3:GetBucket*",
-#         "s3:ListBucket",
-#         "s3:PutObject",
-#         "s3:PutObjectAcl",
-#         "s3:PutBucketNotification"
-#     ]
-#     resources = ["*"]
-#   }
-#   statement {
-#     principals {
-#       type        = "Service"
-#       identifiers = ["s3.amazonaws.com"]
-#     }
-#     actions = [
-#         "s3:PutObject"
-#     ]
-#     resources = ["arn:aws:s3:::orca-archive-worm/*"]
-#   }
-# }
-
-# resource "aws_s3_bucket_policy" "orca-reports-policy" {
-#   bucket = aws_s3_bucket.orca-reports.id
-#   policy = data.aws_iam_policy_document.orca-reports-policy-json.json
-# }
+data "aws_iam_policy_document" "orca-reports-policy-json" {
+  statement {
+    principals {
+      type        = "AWS"
+      identifiers = ["arn:aws:iam::236859827343:root"]
+    }
+    actions = [
+        "s3:GetObject*",
+        "s3:GetBucket*",
+        "s3:ListBucket",
+        "s3:PutObject",
+        "s3:PutObjectAcl",
+        "s3:PutBucketNotification"
+    ]
+    resources = [
+        "arn:aws:s3:::PREFIX-orca-reports",
+        "arn:aws:s3:::PREFIX-orca-reports/*"
+        ]
+  } 
+  statement {
+    principals {
+      type        = "Service"
+      identifiers = ["s3.amazonaws.com"]
+      }
+      actions = [
+                  "s3:PutObject"
+      ]
+      resources = [
+                  "arn:aws:s3:::PREFIX-orca-reports",
+                  "arn:aws:s3:::PREFIX-orca-reports/*"
+                  ]
+      condition {
+        test     = "StringEquals"
+        variable = "s3:x-amz-acl"
+        values = [
+          "bucket-owner-full-control"
+        ]
+      }
+      condition {
+        test     = "StringEquals"
+        variable = "aws:SourceAccount"
+        values   = ["782417781503"]
+      }
+      condition {
+        test = "ArnLike"
+        variable = "aws:SourceArn"
+        values = [
+          "arn:aws:s3:::PREFIX-*"
+        ]
+      }
+    }
+  statement {
+    principals {
+      type        = "Service"
+      identifiers = ["s3.amazonaws.com"]
+      }
+      actions = [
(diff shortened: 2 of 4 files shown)


website/docs/developer/development-guide/code/versioning-releases.md
```diff
@@ -114,28 +114,30 @@ For testing, use your feature branch in cumulus-orca github repo and `ORCA-test-
...
+
 - `<PREFIX>-internal`
 - `<PREFIX>-level0`
 - `<PREFIX>-public`
 - `<PREFIX>-private`
 - `<PREFIX>-protected`
 - `<PREFIX>-orca-primary`
 - `<PREFIX>-tf-state` (for storing the terraform state file)
+- `<PREFIX>-orca-archive-worm`
+- `<PREFIX>-orca-reports`
```
AndrewDorn, line 134: There is no differentiation in this list between the buckets that are automatically created, and the buckets that must be manually created. Recommend having info on the DR OU and Cumulus OU buckets in separate steps.

website/docs/developer/development-guide/code/versioning-releases.md
```diff
@@ -114,28 +114,30 @@ For testing, use your feature branch in cumulus-orca github repo and `ORCA-test-
...
 
-Note that the `<PREFIX>-orca-primary` bucket should be created in the DR OU if we want a full test. This will also need a cross account policy applied to it as well which should be similar to this [policy](https://github.com/nasa/cumulus-orca/blob/develop/website/docs/developer/research/research-s3-bucket-policies.md#reports-bucket).
+Note that the `<PREFIX>-orca-primary`, `<PREFIX>-orca-archive-worm` and `<PREFIX>-orca-reports` buckets should be created in the DR OU if we want a full test. This will also need a cross account policy applied to it as well which should be similar to this [policy](https://github.com/nasa/cumulus-orca/blob/develop/website/docs/developer/research/research-s3-bucket-policies.md#reports-bucket).
```
AndrewDorn, line 136: Group this just with the DR OU bucket optional step.

AndrewDorn (changes requested):
A couple documentation simplifications.

website/docs/developer/development-guide/code/versioning-releases.md
rizbihassan: currently, all the buckets are automatically created. 

Updated the doc to separate/organize the bucket deployment steps.

integration_test/deploy-dr-buckets.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+set -ex
+
+export AWS_ACCESS_KEY_ID=$bamboo_AWS_ACCESS_KEY_ID
```
bhazuka, line 4: Since everything is in the same plan currently, won't using the same variables `$bamboo_AWS_ACCESS_KEY_ID`, `$bamboo_AWS_SECRET_ACCESS_KEY` cause us issues since they would point to the Cumulus OU and not the DR OU? It may be better to call then out as the DR Key and secret. Thoughts?

integration_test/deploy-dr-buckets.sh
```diff
@@ -0,0 +1,27 @@
...
+    echo "Terraform state file found in S3 bucket. Using S3 remote backend"
+fi
+
+terraform init -input=false
+echo "Deploying S3  buckets in Disaaster Recovery account"
+terraform apply \
+  -auto-approve \
+  -input=false
+
+#copy terraform state file to the created tf-state bucket
```
bhazuka, line 26: Why do we make the copy? Shouldn't that be configured beforehand? If we have a chicken and egg problem maybe we should look at https://github.com/stavxyz/terraform-aws-backend to create our state bucket and dynamo table.

integration_test/deploy-dr-buckets.sh
```diff
@@ -0,0 +1,27 @@
...
+export AWS_SECRET_ACCESS_KEY=$bamboo_AWS_SECRET_ACCESS_KEY
+export AWS_DEFAULT_REGION=$bamboo_AWS_DEFAULT_REGION
+
+#remove old files from bamboo as they throw error
+rm *.tf
+git clone --branch develop --single-branch https://github.com/nasa/cumulus-orca.git && cd integration_test
+#replace prefix with bamboo prefix variable
+sed -e 's/PREFIX/'"$bamboo_PREFIX"'/g' dr-buckets.tf.template > dr-buckets.tf
+
+if ! aws s3 cp s3://$bamboo_PREFIX-dr-tf-state/dr-buckets-tf/terraform.tfstate .;then
```
bhazuka, line 14: Are we checking for a file here or the bucket, or both? Not sure we even really need this check since with terraform we should be able to run over and over again.

website/docs/developer/development-guide/code/versioning-releases.md
```diff
@@ -108,30 +108,39 @@ To delete a published tag to re-tag, follow these steps:
...
+
+- AWS_ACCESS_KEY_ID(for cumulus sandbox account)
+- AWS_SECRET_ACCESS_KEY(for cumulus sandbox account)
+- PREFIX
+- AWS_ACCOUNT_ID(for cumulus sandbox account)
+- DB_ADMIN_PASSWORD
+- DB_USER_PASSWORDThese are the buckets that need to be created in cumulus OU account:
+
+
+These are the optional buckets that can be created in disaster recovery AWS account
```
bhazuka, line 134: remove the word 'optional'.

website/docs/developer/development-guide/code/versioning-releases.md
bhazuka: Per @AndrewDorn  comment this section is still a little confusing to me  and I feel we could better separate this to make it clear.

bhazuka (changes requested):
Couple of questions and thoughts.

website/docs/developer/development-guide/code/versioning-releases.md
rizbihassan: Yes updated the paragraphs

integration_test/deploy-dr-buckets.sh
rizbihassan: It will not throw an error since we have to override those variables while running the stages in bamboo but might look a bit confusing.

I will create new different variables for aws access keys in DR account in bamboo to make it less confusing

integration_test/deploy-dr-buckets.sh
rizbihassan: we are checking for the terraform.tf state file. 

When a user runs the pipeline for the first time, the script will fail without this check because there is no .tfstate file or prefix-tf-state bucket. So if it cannot copy the state file, then it means the prefix-tf-state bucket is not created and will deploy all other buckets.

Next time the user reruns the pipeline, this check will find the .tfstate file and only update if there is a modification in terraform resource.

integration_test/deploy-dr-buckets.sh
rizbihassan: we need a `prefix-tf-state` bucket created first before we can use that for configuring as backend. 

so first time when the script is run, it copies the state file to the created bucket to use as backend. 

I did go through the link you sent but found that somewhat confusing to use. Plus seems like they have some additional bucket properties we don't need and dynamodb properties are different https://github.com/stavxyz/terraform-aws-backend/blob/master/main.tf

The https://github.com/stavxyz/terraform-aws-backend/blob/beb7ee0b229701f47c9853fb30529dbc568da0e5/main.tf#L83 will also cause issue while deleting the resource later on

If we decide to create the tf-state bucket and dynamodb table manually first instead of automating, then we will not have this dependency.

website/docs/developer/development-guide/code/versioning-releases.md
AndrewDorn: They're all upgraded to automatic deployment? Where is, say, the level0 bucket created?

rizbihassan: yes correct. Level0 is created in cumulus sandbox account. 
See this commit 
https://github.com/nasa/cumulus-orca/pull/297/commits/c805b131cf00f228a71c93c5bc2359b2079b12e7

website/docs/developer/development-guide/code/versioning-releases.md
```diff
@@ -108,30 +108,39 @@ To delete a published tag to re-tag, follow these steps:
...
-
-- AWS_ACCESS_KEY_ID
-- AWS_SECRET_ACCESS_KEY
-- PREFIX
-- AWS_ACCOUNT_ID
-- DB_ADMIN_PASSWORD
-- DB_USER_PASSWORD
-
-The ORCA buckets and dynamoDB table are created automatically in bamboo `Deploy Dev Cumulus and ORCA Stack` stage. However, these can also be created manually if desired by the user. These are the buckets that need to be created:
+The ORCA buckets and dynamoDB table in cumulus OU account are created automatically in bamboo `Deploy Dev Cumulus and ORCA Stack` stage. These are the buckets that need to be created in cumulus OU account:
```
AndrewDorn, line 115: "The Cumulus and TF buckets"

website/docs/developer/development-guide/code/versioning-releases.md
```diff
@@ -108,30 +108,39 @@ To delete a published tag to re-tag, follow these steps:
...
+
+- CUMULUS_AWS_ACCESS_KEY_ID
+- CUMULUS_AWS_SECRET_ACCESS_KEY
+- PREFIX
+- AWS_ACCOUNT_ID(for cumulus sandbox account)
+- DB_ADMIN_PASSWORD
+- DB_USER_PASSWORDThese are the buckets that need to be created in cumulus OU account:
+
+
+These are the buckets that can be created in disaster recovery AWS account
```
AndrewDorn, line 134: "These are the Orca buckets that are created"

website/docs/developer/development-guide/code/versioning-releases.md
```diff
@@ -108,30 +108,39 @@ To delete a published tag to re-tag, follow these steps:
...
+- `<PREFIX>-tf-state` (for storing the terraform state file in cumulus OU account)
+
+While running the `Deploy Dev RDS Stack` stage, replace the following variables with yours. This is because some variables are sensitive and some will vary depending upon the user running the pipeline. Hitting 'play' next to `Deploy DR ORCA buckets`, `Deploy Dev RDS Stack` and `Deploy Dev Cumulus and ORCA Stack` brings up a checkbox list to run multiple jobs at once. Note that none of the checkboxes should be checked.
+
+- CUMULUS_AWS_ACCESS_KEY_ID
+- CUMULUS_AWS_SECRET_ACCESS_KEY
+- PREFIX
+- AWS_ACCOUNT_ID(for cumulus sandbox account)
+- DB_ADMIN_PASSWORD
+- DB_USER_PASSWORDThese are the buckets that need to be created in cumulus OU account:
```
AndrewDorn, line 131: Some extra words here

website/docs/developer/development-guide/code/versioning-releases.md
rizbihassan: yes it was created in the previous PR. This template creates the buckets https://github.com/nasa/cumulus-orca/blob/develop/integration_test/buckets.tf.template

website/docs/developer/development-guide/code/versioning-releases.md
rizbihassan: changed

AndrewDorn (approved):

bhazuka (changes requested):
I feel like the state management is not quite correct.

integration_test/deploy-dr-buckets.sh
rizbihassan: In the first check, if it's not able to copy the .tfstate file from the s3 bucket, it means the file is not present and the bucket is not created. If the check is able to copy the tfstate file, it means the bucket is already created and then it configures the s3 backend. 

If we want to manually create the bucket for tf state first, then configuring the backend will be easier.

bhazuka: Lets talk through this during orcatechture or after standup tomorrow.

integration_test/dr-buckets.tf.template
```diff
@@ -0,0 +1,154 @@
...
+      actions = [
+                  "s3:PutObject"
+      ]
+      resources = [
+                  "arn:aws:s3:::PREFIX-orca-reports",
+                  "arn:aws:s3:::PREFIX-orca-reports/*"
+                  ]
+      condition {
+        test     = "StringEquals"
+        variable = "s3:x-amz-acl"
```
AndrewDorn, line 99: You can add in your storage-class condition where appropriate.

AndrewDorn (approved):
Might want to add in the storage-class stuff to stick close to real buckets, but otherwise looks good.

integration_test/dr-buckets.tf.template
rizbihassan: shouldn't that policy be attached to only the glacier archive bucket that we currently don't have in terraform?

integration_test/deploy-dr-buckets.sh
```diff
@@ -0,0 +1,48 @@
...
+export AWS_SECRET_ACCESS_KEY=$bamboo_DR_AWS_SECRET_ACCESS_KEY
+export AWS_DEFAULT_REGION=$bamboo_AWS_DEFAULT_REGION
+
+#remove old files from bamboo as they throw error
+rm *.tf
+git clone --branch develop --single-branch https://github.com/nasa/cumulus-orca.git && cd integration_test
+#replace prefix with bamboo prefix variable
+sed -e 's/PREFIX/'"$bamboo_PREFIX"'/g' dr-buckets.tf.template > dr-buckets.tf
+
+if ! aws s3api head-bucket --bucket $bamboo_PREFIX-dr-tf-state;then
```
bhazuka, line 14: You may want to wrap your variable for better readability. So `$bamboo_PREFIX-dr-tf-state` becomes `${bamboo_PREFIX}-dr-tf-state` That would mean if `$bamboo_PREFIX=myvariable` your bucket name would be `myvariable-dr-tf-state` In addition, you may want to make sure the `$bamboo_PREFIX` variable is populated and either set a sane default or error completely if it is not set.

integration_test/dr-buckets.tf.template
```diff
@@ -0,0 +1,154 @@
...
+  tags = {
+    Deployment = "PREFIX"
+    Application = "ORCA"
+  }
+}
+data "aws_iam_policy_document" "orca-primary-policy-json" {
+ statement {
+   principals {
+     type        = "AWS"
+     identifiers = ["arn:aws:iam::236859827343:root"]
```
bhazuka, line 135: Should limit to GLACIER and DEEP_GLACIER for uploads per your previous ticket.

integration_test/dr-buckets.tf.template
```diff
@@ -0,0 +1,154 @@
+resource "aws_s3_bucket" "orca-archive-worm" {
```
bhazuka, line 1: I am not seeing where you are creating a policy for the WORM bucket. I would imagine we would need one similar to the primary with some tweaks.

bhazuka (changes requested):
Looking better. Still a few things to address.

integration_test/dr-buckets.tf.template
rizbihassan: added

integration_test/dr-buckets.tf.template
rizbihassan: added versioning to the bucket according to bucket policy instructions webpage.

Are you also suggesting S3 uploads should limit to GLACIER and DEEP_GLACIER similar to orca-primary bucket?

integration_test/dr-buckets.tf.template
rizbihassan: added to orca-primary

integration_test/dr-buckets.tf.template
rizbihassan: will add policy for s3 uploads and s3 object lock

rizbihassan: added

integration_test/dr-buckets.tf.template
```diff
@@ -1,17 +1,58 @@
...
     status = "Enabled"
   }
 }
+
+resource "aws_s3_bucket_object_lock_configuration" "orca-archive-worm-object-lock-config" {
+  bucket = aws_s3_bucket.orca-archive-worm.bucket
+  rule {
+    default_retention {
+      mode = "GOVERNANCE"
+      days = 180
```
rizbihassan, line 23: kept it as 6 months according to @AndrewDorn suggestion in his research doc:
"With a default policy of Governance mode and an arbitrarily long Retention period, data will be protected from accidental deletion by general users."

PR #310, opened 2022-08-04
added jq package to docker image

rizbihassan:
## Summary of Changes

Added jq package needed for testing 
## Changes

* updated base_images/Dockerfile.bamboo
* updated base_images/Dockerfile.local

## Type of change

### DIFF
--- a/base_images/Dockerfile.bamboo
+++ b/base_images/Dockerfile.bamboo
@@ -17,7 +17,7 @@ yum update -y
 # install Python 3, pip, CLI utilities, session manager and other tools
 RUN \
 yum install -y gcc git make openssl wget zip unzip tar python3-devel nodejs yarn python3-pip awscli session-manager-plugin.rpm procps parallel && \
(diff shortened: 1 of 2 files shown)


AndrewDorn (approved):

base_images/Dockerfile.bamboo
```diff
@@ -17,7 +17,7 @@ yum update -y
 # install Python 3, pip, CLI utilities, session manager and other tools
 RUN \
 yum install -y gcc git make openssl wget zip unzip tar python3-devel nodejs yarn python3-pip awscli session-manager-plugin.rpm procps parallel && \
-pip3 install -q --upgrade pip --trusted-host pypi.org --trusted-host files.pythonhosted.org pydoc-markdown bandit safety flake8 black isort
+pip3 install -q --upgrade pip --trusted-host pypi.org --trusted-host files.pythonhosted.org pydoc-markdown bandit safety flake8 black isort jq
```
bhazuka, line 20: jq is also a linux binary and can be installed by adding it to the `yum install` line. You may want to think about installing it via yum instead of pip.

bhazuka (approved):
Another option for jq

PR #317, opened 2022-08-16
Update KMS key and IAM role policy to add describe action to the key

rizbihassan:
## Summary of Changes

Addresses [ORCA-515: Update KMS key and IAM role policy to add describe action to the key](https://bugs.earthdata.nasa.gov/browse/ORCA-515)

## Changes

* added kms:DescribeKey to iam role and kms key policy

## Type of change

## How Has This Been Tested?

deployed in aws.

## Checklist:

    - [ x] Snyk
    - [x ] git-secrets

### DIFF
--- a/modules/secretsmanager/main.tf
+++ b/modules/secretsmanager/main.tf
@@ -84,7 +84,7 @@ data "aws_iam_policy_document" "orca_kms_key_policy" {
 
     principals {
       type        = "AWS"
-      identifiers = local.iam_users
+      identifiers = local.iam_users 
     }
   }
   statement {
--- a/modules/secretsmanager/main.tf
+++ b/modules/secretsmanager/main.tf
@@ -4,8 +4,6 @@
 ## SECRET MANAGER DATA LOOKUPS
 ## =============================================================================
 
-data "aws_iam_users" "orca_users" {}
-
 data "aws_caller_identity" "current_account" {}
 
 data "aws_region" "current_region" {}
@@ -13,7 +11,6 @@ data "aws_region" "current_region" {}
 
 ## Local Variables
 locals {
-  iam_users    = data.aws_iam_users.orca_users.arns
   account_id   = data.aws_caller_identity.current_account.account_id
   region       = data.aws_region.current_region.name
   kms_arn    = "arn:aws:kms:${local.region}:${local.account_id}:key/*"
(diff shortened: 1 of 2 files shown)


AndrewDorn (approved):
Add a comment that indicates this is a request from Cumulus, and I think we're good.

rizbihassan:
seems like the changes above are not working for podaac 
https://eosdis.slack.com/archives/CSXEM5MPX/p1660320565600179

This could likely be something NGAP should look into as well

rizbihassan:
KMS issue faced by 
- Jenny Liu in version 4.0.1
- Eshin Yen (PODAAC) version 4.0.3

manual steps used to get the deployment working 

```
$ terraform state rm "module.orca.module.orca.module.orca_secretsmanager.aws_kms_key.orca_kms_key
$ terraform state rm module.orca.module.orca.module.orca_secretsmanager
$ aws secretsmanager delete-secret --secret-id <PREFIX>-orca-db-login-secret --force-delete-without-recovery
```

Then terraform apply worked for other teams.

modules/secretsmanager/main.tf
```diff
@@ -84,14 +84,15 @@ data "aws_iam_policy_document" "orca_kms_key_policy" {
 
     principals {
       type        = "AWS"
-      identifiers = local.iam_users
+      identifiers = local.iam_users
```
rizbihassan, line 87: Should we make it more open by changing identifier to root instead of orca_users? Not sure if this is also causing the kms issue
```
identifiers = ["arn:aws:iam::${data.aws_caller_identity.current_account.account_id}:root"]
```

AndrewDorn (changes requested):
I meant comment in the code, so we know it isn't an internal requirement when we're reviewing/maintaining.

rizbihassan:
> I meant comment in the code, so we know it isn't an internal requirement when we're reviewing/maintaining.

yes understood. Will add that as well

modules/secretsmanager/main.tf
bhazuka: I believe Nate did this to help with our security posture and limit access to the keys.

bhazuka (approved):

modules/secretsmanager/main.tf
rizbihassan: talked with Nate and decided to open the policy like above for now

AndrewDorn (approved):

PR #335, opened 2022-09-23
Feature/orca 538 Remove "glacier" bucket name in TF

AndrewDorn:
## Summary of Changes

Removed a few remaining references to "glaicer" bucket.

Addresses [ORCA-538: Remove "glacier" bucket name in TF](https://bugs.earthdata.nasa.gov/browse/ORCA-538)

## Changes

* Removed references to "glacier" bucket.
* Fixed bug where if request_files_from_archive was run on its own, default recovery type would not be applied.

## Type of change

## How Has This Been Tested?

Tests pass locally

## Checklist:

    - [x] User documentation
    - [x] Unit tests
    - [x] Snyk
    - [x] git-secrets

### DIFF
--- a/tasks/request_from_archive/API.md
+++ b/tasks/request_from_archive/API.md
@@ -3,7 +3,7 @@
 * [request\_from\_archive](#request_from_archive)
   * [RestoreRequestError](#request_from_archive.RestoreRequestError)
   * [task](#request_from_archive.task)
-  * [get\_glacier\_recovery\_type](#request_from_archive.get_glacier_recovery_type)
+  * [get\_archive\_recovery\_type](#request_from_archive.get_archive_recovery_type)
   * [inner\_task](#request_from_archive.inner_task)
   * [process\_granule](#request_from_archive.process_granule)
   * [get\_s3\_object\_information](#request_from_archive.get_s3_object_information)
@@ -15,7 +15,7 @@
 # request\_from\_archive
 
 Name: request_from_archive.py
-Description:  Lambda function that makes a restore request from glacier for each input file.
+Description:  Lambda function that makes a restore request from archive bucket for each input file.
 
 <a id="request_from_archive.RestoreRequestError"></a>
 
@@ -57,16 +57,17 @@ then calls inner_task.
 
 - `RestoreRequestError` - Thrown if there are errors with the input request.
 
-<a id="request_from_archive.get_glacier_recovery_type"></a>
+<a id="request_from_archive.get_archive_recovery_type"></a>
 
-#### get\_glacier\_recovery\_type
+#### get\_archive\_recovery\_type
 
 ```python
-def get_glacier_recovery_type(config: Dict[str, Any]) -> str
+def get_archive_recovery_type(config: Dict[str, Any]) -> str
 ```
 
-Returns the glacier recovery type from either config or environment variable.
+Returns the archive recovery type from either config or environment variable.
 Must be either 'Bulk', 'Expedited', or 'Standard'.
+Defaults to 'Standard' if none found.
 
 **Arguments**:
 
@@ -95,7 +96,7 @@ if it fails, waiting {retry_sleep_secs} between each attempt.
   this may not directly correspond to Lambda input.
 - `event` - A dict with the following keys:
 - `'config'` _dict_ - A dict with the following keys:
-- `'defaultBucketOverride'` _str_ - The name of the glacier bucket
+- `'defaultBucketOverride'` _str_ - The name of the archive bucket
   from which the files will be restored.
 - `'asyncOperationId'` _str_ - The unique identifier used for tracking requests.
 - `'input'` _dict_ - A dict with the following keys:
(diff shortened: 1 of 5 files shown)


tasks/request_from_archive/API.md
```diff
@@ -174,42 +175,39 @@ Perform a head request to get information about a file in S3.
...
-def restore_object(s3_cli: BaseClient,
-                   key: str,
-                   days: int,
-                   db_glacier_bucket_key: str,
-                   attempt: int,
-                   job_id: str,
-                   recovery_type: str = "Standard") -> None
+def restore_object(s3_cli: BaseClient, key: str, days: int,
+                   db_archive_bucket_key: str, attempt: int, job_id: str,
+                   recovery_type: str) -> None
```
AndrewDorn, line 195: Since this is always called with the parameter present, the default was never used.

tasks/request_from_archive/request_from_archive.py
```diff
@@ -556,30 +564,30 @@ def restore_object(
     s3_cli: BaseClient,
     key: str,
     days: int,
-    db_glacier_bucket_key: str,
+    db_archive_bucket_key: str,
     attempt: int,
     job_id: str,
-    recovery_type: str = "Standard",
```
AndrewDorn, line 562: Since this is always called with the parameter present, the default was never used.

tasks/request_from_archive/request_from_archive.py
```diff
@@ -166,12 +167,13 @@ def get_glacier_recovery_type(config: Dict[str, Any]) -> str:
...
     # Look for config override
     recovery_type = config.get(CONFIG_DEFAULT_RECOVERY_TYPE_OVERRIDE_KEY, None)
-    if recovery_type is not None:
+    if recovery_type is not None:  # either return or raise error.
         if recovery_type in VALID_RESTORE_TYPES:
             LOGGER.info(
                 f"Using restore type of {recovery_type} "
                 f"found in the configuration {CONFIG_DEFAULT_RECOVERY_TYPE_OVERRIDE_KEY} key."
             )
+            return recovery_type
```
AndrewDorn, line 176: Halts processing once override is found, reducing complexity.

tasks/request_from_archive/request_from_archive.py
```diff
@@ -183,35 +185,40 @@ def get_glacier_recovery_type(config: Dict[str, Any]) -> str:
...
-            else:
-                LOGGER.error(
-                    f"Invalid restore type value of '{recovery_type}' "
-                    f"found in environment variable {OS_ENVIRON_DEFAULT_RECOVERY_TYPE_KEY}."
-                )
-                raise ValueError(
-                    f"Invalid restore type value in environment variable "
-                    f"{OS_ENVIRON_DEFAULT_RECOVERY_TYPE_KEY}"
-                )
+        recovery_type = "Standard"
```
AndrewDorn, line 190: Default now set here. Since TF will always set the os.environ default, this did not affect customers.

tasks/request_from_archive/request_from_archive.py
```diff
@@ -167,51 +168,43 @@ def get_glacier_recovery_type(config: Dict[str, Any]) -> str:
...
             )
         else:
-            LOGGER.error(
-                f"Invalid restore type value of '{recovery_type}' "
-                f"found in the configuration {CONFIG_DEFAULT_RECOVERY_TYPE_OVERRIDE_KEY} key."
-            )
-            raise ValueError(
-                f"Invalid restore type value in configuration "
-                f"{CONFIG_DEFAULT_RECOVERY_TYPE_OVERRIDE_KEY} key."
+            recovery_type = "Standard"
```
AndrewDorn, line 184: Default now set here. Since TF will always set the os.environ default, this did not affect customers.

tasks/request_from_archive/request_from_archive.py
```diff
@@ -167,51 +168,43 @@ def get_glacier_recovery_type(config: Dict[str, Any]) -> str:
...
-                LOGGER.error(
-                    f"Invalid restore type value of '{recovery_type}' "
-                    f"found in environment variable {OS_ENVIRON_DEFAULT_RECOVERY_TYPE_KEY}."
-                )
-                raise ValueError(
-                    f"Invalid restore type value in environment variable "
-                    f"{OS_ENVIRON_DEFAULT_RECOVERY_TYPE_KEY}"
-                )
-
+    if recovery_type not in VALID_RESTORE_TYPES:
```
AndrewDorn, line 190: Moved ValueError check to end. Will now need to be combined with info logging to fully explain error, but reduces complexity.

tasks/request_from_archive/request_from_archive.py
bhazuka: Instead of setting a value of `recovery_type = "Standard"` here I wonder if we shouldn't raise a `ValueError` here? In order to trigger this, that would mean, someone would have to purposefully delete the environment key from the lambda or set the TF value to None. I feel like we should throw an error and exit in that case instead of setting a default value of `Standard`. Thoughts?

tasks/request_from_archive/request_from_archive.py
bhazuka: I think this is smart since we only need to validate that the value is correct once when we choose where we get the value from override or default.

tasks/request_from_archive/request_from_archive.py
bhazuka: Concur, I also feel this value should never be None and should always be set.

bhazuka (changes requested):
Looks good overall. A few thoughts questions on setting the default recovery type to comment on.

bhazuka (approved):

PR #354, opened 2022-11-10
Feature/orca 579 Research and prototype load balancers for GraphQL

AndrewDorn:
## Summary of Changes

Added Application Load Balancer that can be used to query GraphQL service.

Addresses [ORCA-557: Develop amazing new feature](https://bugs.earthdata.nasa.gov/browse/ORCA-557)

## Changes

* Added Application Load Balancer to TF.
* Fixed some bugs with GQL security setup.

## Type of change

## How Has This Been Tested?

Deployed to AWS and accessed via lambda in same VPC.

## Checklist:

    - [x] CHANGELOG.md
    - [x] Snyk
    - [x] git-secrets

### DIFF
--- a/modules/graph_ql/main.tf
+++ b/modules/graph_ql/main.tf
@@ -1,6 +1,6 @@
-# load balancer
-resource "aws_lb" "graphql_load_balancer" {
-  name               = "${var.prefix}-graphql-load-balancer"
+# Application load balancer
+resource "aws_lb" "gql_app_lb" {
+  name               = "${var.prefix}-gql-app-lb"
   internal           = true
   load_balancer_type = "application"
   security_groups    = [var.vpc_postgres_ingress_all_egress_id]
@@ -9,8 +9,8 @@ resource "aws_lb" "graphql_load_balancer" {
   tags               = var.tags
 }
 
-resource "aws_lb_target_group" "graphql_load_balancer_target_group" {
-  name        = "${var.prefix}-graphql-target-group"
+resource "aws_lb_target_group" "gql_app_lb_target_group" {
+  name        = "${var.prefix}-gql-app-lb-t"
   vpc_id      = var.vpc_id
   protocol    = "HTTP"
   port        = 5000
@@ -26,8 +26,19 @@ resource "aws_lb_target_group" "graphql_load_balancer_target_group" {
   tags = var.tags
 }
 
+resource "aws_lb_listener" "gql_app_lb_listener" {
+  load_balancer_arn = aws_lb.gql_app_lb.arn
+  port              = "5000"
+  protocol          = "HTTP"
+
+  default_action {
+    type             = "forward"
+    target_group_arn = aws_lb_target_group.gql_app_lb_target_group.arn
+  }
+}
+
 # ecs service and task
-data "aws_iam_policy_document" "graphql_task_execution_policy_document" {
+data "aws_iam_policy_document" "gql_task_execution_policy_document" {
   statement {
     actions   = ["sts:AssumeRole"]
     resources = [aws_iam_role.orca_ecs_tasks_role.arn]
@@ -61,10 +72,10 @@ resource "aws_iam_role" "orca_ecs_tasks_role" {
   tags                 = var.tags
 }
 
-resource "aws_iam_role_policy" "graphql_task_role_policy" {
-  name   = "${var.prefix}_orca_graphql_task_role_policy"
+resource "aws_iam_role_policy" "gql_task_role_policy" {
+  name   = "${var.prefix}_orca_gql_task_role_policy"
   role   = aws_iam_role.orca_ecs_task_execution_role.id
-  policy = data.aws_iam_policy_document.graphql_task_execution_policy_document.json
+  policy = data.aws_iam_policy_document.gql_task_execution_policy_document.json
 }
 
 resource "aws_iam_role" "orca_ecs_task_execution_role" {
@@ -75,8 +86,8 @@ resource "aws_iam_role" "orca_ecs_task_execution_role" {
 }
 
 # Defines how the image will be run.
-resource "aws_ecs_task_definition" "graphql_task" {
-  family                   = "${var.prefix}_orca_graphql_task"
+resource "aws_ecs_task_definition" "gql_task" {
+  family                   = "${var.prefix}_orca_gql_task"
   network_mode             = "awsvpc"
   requires_compatibilities = ["FARGATE"]
   cpu                      = "1024"
@@ -87,7 +98,7 @@ resource "aws_ecs_task_definition" "graphql_task" {
   container_definitions    = <<DEFINITION
 [
   {
-    "name": "orca-graphql",
+    "name": "orca-gql",
     "image": "ghcr.io/nasa/cumulus-orca/graphql:0.0.18",
     "cpu": 512,
     "memory": 256,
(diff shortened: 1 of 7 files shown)


modules/graph_ql/output.tf
```diff
@@ -0,0 +1,4 @@
+output "graphql_load_balancer_dns_name" {
+  value       = data.aws_lb.gql_app_lb_data.dns_name
+  description = "The DNS Name of the Application Load Balancer that handles access to ORCA GraphQL."
+}
```
AndrewDorn, line 4: Felt it was important to show how to output this, but didn't want to bubble it up to a customer-visible layer. Could change/remove this if desired.

modules/graph_ql/main.tf
```diff
@@ -1,4 +1,133 @@
...
+
+  default_action {
+    type             = "forward"
+    target_group_arn = aws_lb_target_group.gql_app_lb_target_group.arn
+  }
+
+  tags               = var.tags
+}
+
+# # Network load balancer
```
AndrewDorn, line 84: Since this is just a tentative solution, any objections to me leaving this code here until we're sure it won't be needed?

modules/graph_ql/main.tf
```diff
@@ -1,4 +1,133 @@
-data "aws_iam_policy_document" "graphql_task_execution_policy_document" {
+# Local Variables
+locals {
+  graphql_port = 5000
```
AndrewDorn, line 3: Our Dockerfile needs a few tweaks to make this truly configurable. Created https://bugs.earthdata.nasa.gov/browse/ORCA-581

modules/graph_ql/output.tf
bhazuka: I feel this is good to keep as we may want to leverage this later for some internal prototyping.

modules/graph_ql/main.tf
```diff
@@ -58,12 +187,22 @@ resource "aws_ecs_task_definition" "graphql_task" {
...
+      {
+        "containerPort": ${local.graphql_port},
+        "hostPort": ${local.graphql_port}
+      }
+    ],
     "environment": [
+      {
+        "name": "PORT",
+        "value": "${local.graphql_port}"
+      }
```
bhazuka, line 205: We would likely want to set the `ENV` value here to from a terraform variable that we would default to the Production instance and be able to override with a "Development" instance.

modules/graph_ql/main.tf
bhazuka: Yes, ok with me. I would add comments though saying this is left in for further testing/study of VPC link, etc. so we do not accidentally remove it without good cause.

modules/graph_ql/main.tf
```diff
@@ -1,4 +1,133 @@
...
+  idle_timeout       = 30 # API Gateway locks us to 30 seconds.
+  tags               = var.tags
+}
+
+data "aws_lb" "gql_app_lb_data" {
+  arn = aws_lb.gql_app_lb.arn
+}
+
+resource "aws_lb_target_group" "gql_app_lb_target_group" {
+  name        = "${random_id.lb_name.hex}-gql-a" # name must be randomized. Max 32 characters. Some prefixes are 25 characters long.
```
bhazuka, line 51: Not seeing where we are adding the PREFIX to the name here with the random ID. Is that intentional or just the way random works?

modules/graph_ql/main.tf
```diff
@@ -1,4 +1,133 @@
...
+  name        = "${var.prefix}-gql"
+  description = "Allow inbound communication on container port."
+  vpc_id      = var.vpc_id
+
+  ingress {
+    description      = "Container port communication."
+    from_port        = local.graphql_port
+    to_port          = local.graphql_port
+    protocol         = "tcp"
+    cidr_blocks      = ["0.0.0.0/0"]
```
bhazuka, line 20: We may want to put a card on the backlog for security research that will
- Check with security on allowing all ingress on a port from anywhere (we may have to limit this to our subnets) and provide an implementation plan if we have to limit it.
- Check with security to see if we need to move HTTP to HTTPS and what those implementation details may look like if we need to. (My hope is since it is within the private VPC we can keep it as HTTP)

modules/graph_ql/main.tf
bhazuka, line 3: A comment here may be good that lets us know that it is used for the container, host, and ALB ports.

bhazuka (changes requested):
Overall looks good, a couple of thoughts and comments for some minor things.

modules/graph_ql/main.tf
AndrewDorn: Yeah, should've added that. Fixed.

modules/graph_ql/main.tf
AndrewDorn: It's more that we literally don't have room for it, given how long customer prefixes are. I don't see a better solution than relying on tags.

modules/graph_ql/main.tf
```diff
@@ -1,4 +1,133 @@
...
+    path = "/healthz"
+  }
+
+  tags = var.tags
+}
+
+resource "aws_lb_listener" "gql_app_lb_listener" {
+  load_balancer_arn = aws_lb.gql_app_lb.arn
+  port              = local.graphql_port
+  protocol          = "HTTP"
```
rizbihassan, line 75: I think `HTTPS ` should be more secured.

modules/graph_ql/main.tf
AndrewDorn: Added

modules/graph_ql/main.tf
```diff
@@ -1,4 +1,133 @@
...
+    protocol         = "-1"
+    cidr_blocks      = ["0.0.0.0/0"]
+    ipv6_cidr_blocks = ["::/0"]
+  }
+
+  tags = var.tags
+}
+
+# Application load balancer
+resource "aws_lb" "gql_app_lb" {
```
rizbihassan, line 37: we can also enable access logs to track activities if desired.

rizbihassan (changes requested):
looks good. Added some minor suggestions.

modules/graph_ql/main.tf
AndrewDorn: Would it be acceptable to instead change the default `ENV` to `production`? Feels like that should really be our default, with overrides being used when additional leniency is desired for development.

modules/graph_ql/main.tf
AndrewDorn: Agreed. Sadly, there are some additional requirements that come with that. https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_listener#certificate_arn

modules/graph_ql/main.tf
AndrewDorn: Gave it a shot, but looks like it requires additional permissions on the `internal` system bucket. Given the difficulties we had the last time we suggested s3 policy updates, recommend we hold off on this until the system is ready for customer integration/migration.

AndrewDorn: https://bugs.earthdata.nasa.gov/browse/ORCA-583

modules/graph_ql/main.tf
AndrewDorn: Changed default, per discussion.

modules/graph_ql/main.tf
AndrewDorn: Found that I had to open traffic to allow the `aws_ecs_service` to retrieve the image. Created separate security group `gql_task_security_group`.

rizbihassan (approved):

modules/graph_ql/main.tf
bhazuka: After thinking about it, wouldn't we still want to put the environment terraform variable in here so that you could change the behavior on deployment without any change to the underlying terraform? That way, as a developer you could deploy a development version while customers could deploy the production version without any code change for either just a terraform variable change?

bhazuka (changes requested):
Everything looking ok for me with added cards. One question to resolve and I think we are good.

bhazuka (approved):

PR #364, opened 2022-12-15
Feature/orca 590 Research replace `safety` with `snyk`

rizbihassan:
## Summary of Changes

Addresses [ORCA-590: Research replace `safety` with `snyk`](https://bugs.earthdata.nasa.gov/browse/ORCA-590)

## Changes

* removed safety use from orca code
* added snyk test in base bin directory

## Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?
Ran bin/run_tests.sh successfully locally and in bamboo.

## Checklist:

    - [x ] Snyk
    - [ x] git-secrets


bin/run_tests.sh
```diff
@@ -4,6 +4,13 @@ set -e
 base=$(pwd)
 failed=0
 
+# run snyk tests locally. Skip in bamboo due to authentication token requirement.
+echo "Scanning Infrastructure as code with Snyk for potential vulnerabilities..."
+snyk iac test || true
```
rizbihassan, line 9: Currently the tests will continue even if there are errors raised in snyk because the issues can be critical and not straightforward which might require creating additional cards for implementation as seen in snyk issues in the past.

bin/run_tests.sh
```diff
@@ -4,6 +4,13 @@ set -e
 base=$(pwd)
 failed=0
 
+# run snyk tests locally. Skip in bamboo due to authentication token requirement.
```
rizbihassan, line 7: if we also want to run these snyk tests in bamboo, then we have to create a variable for passing the snyk authentication token and the pipeline will have to be run manually which is extra step.
I also noticed there is already a stage thats runs snyk scan in bamboo. 
See https://github.com/nasa/cumulus-orca/blob/develop/bamboo-specs/bamboo.yaml#L96

rizbihassan:
Steps performed for installing snyk cli in mac:

```
brew tap snyk/tap
brew install snyk
```

Once installed, authenticate using:
`snyk auth <TOKEN>`

The token can be found in the Snyk app in browser under "Account Settings", and then "API Token". The token was already created in my case.

bin/run_tests.sh
AndrewDorn: Because I'm me, I lean towards loud errors. We currently hold up PRs for snyk errors, so it feels consistent to raise an error in scripts.

bin/run_tests.sh
AndrewDorn, line 7: What stops this from being run in bamboo? I don't recognize the branch point.

lambda scripts/run_tests.sh
```diff
@@ -72,12 +72,6 @@ bandit -r *.py test
 check_returncode $? "ERROR: Potential security or code issues found."
 
 
-## Check code third party libraries for CVE issues
-echo "INFO: Running checks on third party libraries ..."
-safety check -r requirements.txt -r requirements-dev.txt
-check_returncode $? "ERROR: Potential security issues third party libraries."
-
-
```
AndrewDorn, line 80: Given that snyk has keys you have to use, this feels like a reasonable call.

bin/run_tests.sh
AndrewDorn, line 7: We should have some docs on the authentication token and how to pass it into this script. Probably fine just adding some comments to this file.

AndrewDorn (changes requested):
Good start. Few spots where I'd appreciate some extra docs, and we should have a discussion on how loud we want this check.

bin/run_tests.sh
rizbihassan: if we also want to run these snyk tests in bamboo, then we have to create a variable for passing the snyk authentication token and the pipeline will have to be run manually which is extra step.
I also noticed there is already a stage thats runs snyk scan in bamboo.
See https://github.com/nasa/cumulus-orca/blob/develop/bamboo-specs/bamboo.yaml#L96

AndrewDorn: But we run this script in Bamboo. So when this script is run in Bamboo, what enforces the "Skip in bamboo due to authentication token requirement"? I don't see an if-check or other branch point.

rizbihassan: I think will still look if there is a built-in bamboo variable that I can use and then use if else condition.

bhazuka: Per our conversation, recommend moving tests for website into a separate script we can run manually locally. For tests on IAC we need to follow up with Jon V and check on licensing. I know last month, they were limited to only 20 or so runs a month on IAC checks. We will need to see if this is still true or if it is now unlimited. I also recommend that we keep the check as a separate script that we can run manually. 

Since we already check things in Bamboo via snyk, recommend that we would only run theses locally.

bin/run_tests.sh
bhazuka: Per conversation, need to discuss with Jon V. on how often we can run IAC tests based on their license.

bin/run_tests.sh
```diff
@@ -4,6 +4,13 @@ set -e
...
+echo "Scanning Infrastructure as code with Snyk for potential vulnerabilities..."
+snyk iac test || true
+echo "Scanning open-source code with Snyk for potential vulnerabilities..."
+cd website && snyk test -d || true
+cd -
+
```
bhazuka: Need to research and come up with a way to scan our python requirements. I would recommend doing the check at each individual lambda level per our conversation. Depending on the situation you may either want to recommend a seperate script for each lambda or integrating it into run_test script for each lambda.

Note that the code will also need to be added to the graphql code as well at some point.

bhazuka (changes requested):
Added comments per our discussion.

bin/run_tests.sh
rizbihassan: after discussion with team, we now have a separate script for snyk that will have to be manually run in our local machine.

rizbihassan: after discussion with team, we now have a separate script for snyk that will have to be manually run in our local machine.

rizbihassan: plus we have decided not to run it in bamboo since bamboo already has snyk job added

rizbihassan: added some comments

bin/run_tests.sh
rizbihassan: After some research with snyk for scanning requirements.txt files, it is giving the following error on mac even after trying all solutions available on the web.
```
copy_from_archive rhassan$ snyk test --file=requirements.txt --package-manager=pip --command=python3
Failed to test pip project 
```
Solutions tried: 
https://stackoverflow.com/questions/67110095/snyk-test-returns-failed-to-test-pip-project-error

https://support.snyk.io/hc/en-us/articles/360007419257-Required-packages-missing-when-testing-a-Python-project-with-pip

https://github.com/snyk/cli/issues/730

https://www.freecodecamp.org/news/how-to-find-and-fix-security-vulnerabilities-using-snyk/

rizbihassan: created https://bugs.earthdata.nasa.gov/browse/ORCA-598
`Add snyk scan for all lambda requirements.txt and requirements-dev.txt files`

AndrewDorn (approved):

bhazuka (approved):
Looks good to me

PR #386, opened 2023-02-15
Feature/orca 573 Enforce strong passwords in _validate_password

rizbihassan:
## Summary of Changes

Addresses [ORCA-573: Enforce strong passwords in _validate_password](https://bugs.earthdata.nasa.gov/browse/ORCA-573)

## Changes

- updated orca_shared/database/use_cases/validation.py to include new password requirements
- added relevant unit tests

## Type of change

## How Has This Been Tested?

added unit tests and also deployed to aws.

## Checklist:

    - [x ] Snyk
    - [x ] git-secrets

Password will now have the following requirements:
- at least one upper case letter
- at least one digit
- at least one special character
- minimum length of 12


shared_libraries/orca_shared/database/use_cases/validation.py
```diff
@@ -27,6 +26,18 @@ def _validate_password(password: str, context: str, logger: logging.Logger) -> N
         msg = f"{context} password must be at least 12 characters long."
         logger.critical(msg)
         raise Exception(msg)
+    elif re.search('[0-9]', password) is None:
+        msg = f"{context} password must contain a digit between 0 and 9."
+        logger.critical(msg)
+        raise Exception(msg)
+    elif re.search('[A-Z]', password) is None:
```
AndrewDorn, line 33: I'm tempted to say we should require lower and upper case.

shared_libraries/orca_shared/database/use_cases/validation.py
```diff
@@ -27,6 +26,18 @@ def _validate_password(password: str, context: str, logger: logging.Logger) -> N
...
+        msg = f"{context} password must contain an upper case letter."
+        logger.critical(msg)
+        raise Exception(msg)
+    elif re.search('[$&%+,:;=?.*^@#]', password) is None:
+        msg = f"{context} password must contain a special character."
+        logger.critical(msg)
+        raise Exception(msg)
```
AndrewDorn, line 40: Good list of checks. And db_deploy will run these during deployment, so this will raise errors during deployment as well.

shared_libraries/orca_shared/database/use_cases/validation.py
```diff
@@ -10,7 +10,6 @@ def validate_config(config: PostgresConnectionInfo, logger: logging.Logger) -> N
     validate_postgres_name(config.admin_username, "Admin username", logger)
 
     _validate_password(config.user_password, "User", logger)
```
AndrewDorn, line 12: Ideally we'd cover the db_admin password as well (so when we decouple we already have the checks in-place), but I haven't looked into the Cumulus password requirements, so I don't know if this is feasible.

shared_libraries/orca_shared/database/use_cases/validation.py
AndrewDorn, line 40: Looking good. Just needs migration notes to tell users how to update their orca-user password. If we do wind up adding admin password validation that will need similar migration notes.

AndrewDorn (changes requested):
Good stuff. One extra requirement I saw, some migration notes, and a discussion with Cumulus that may get its own card.

shared_libraries/orca_shared/database/use_cases/validation.py
rizbihassan: yes it will fail. Added steps to migration notes.

shared_libraries/orca_shared/database/use_cases/validation.py
rizbihassan: yup added.

shared_libraries/orca_shared/database/use_cases/validation.py
rizbihassan: added

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and includes an additional section for migration notes.
 
 ## [Unreleased]
 ### Added
+
```
AndrewDorn, line 18: Empty line

CHANGELOG.md
```diff
@@ -24,11 +25,23 @@ and includes an additional section for migration notes.
 - *ORCA-597*
   - Server access logging is now enabled for graphql application load balancer.
 
+### Changed
+- *ORCA-573* Updated ORCA DB user password to now have a stronger password requirement. See migration notes for details.
+
+
```
AndrewDorn, line 31: We typically only have one line of whitespace between sections.

CHANGELOG.md
rizbihassan: removed

CHANGELOG.md
rizbihassan: fixed

CHANGELOG.md
```diff
@@ -24,11 +25,23 @@ and includes an additional section for migration notes.
...
   - API Gateway access is now deprecated, and will be removed in a future update.
   - Use the `orca_graphql_load_balancer_dns_name` variable to send your queries to GraphQL as json strings in a POST request.
+- Users will need to update their orca-user password. The password must have the following requirements otherwise the `db_deploy` lambda will fail during deployment.
+  - one upper case letter
+  - one lower case letter
+  - one digit
+  - one special character
+  - minimum length of 12
+
+  Update `db_user_password` variable in your `cumulus-tf/terraform.tfvars` file to match the new password requirement and then run terraform.
```
AndrewDorn, line 48: I don't think this will be enough, unfortunately. db_deploy doesn't recreate the user account on every TF apply.
Try changing your password and then hitting one of the API Gateway reporting endpoints. My guess is that you'll see a permissions/login error.

AndrewDorn (changes requested):
I have a concern about the migration plan. Unless I'm missing something, may require more steps or a db migration.

CHANGELOG.md
rizbihassan: hmm looks like it throws autentication error
`FATAL:  password authentication failed for user \"rhrh_orcauser`

AndrewDorn: Yup. You'll need something that can reset the password on that user. You can either create manual steps or, my preferred solution, have db_deploy reset that users password. Probably as the first step in this block: https://github.com/nasa/cumulus-orca/blob/ca65aaf37f0738540f4f343407231da42e84fbba/tasks/db_deploy/db_deploy.py#L104
Without trying it myself, I believe the syntax you need is "ALTER ROLE "{user_name}" WITH ENCRYPTED PASSWORD :user_password;" https://www.postgresql.org/docs/current/sql-alterrole.html

rizbihassan: ok I am currently looking into updating db_deploy to reset the password using that `alter role` command above. 
Will let you know the outcome

AndrewDorn: Looked a little closer, and if this plan does work, it would actually go at the top of the the `if app_schema_exists` block. With my initial suggestion you'd have this new code running before the user it modifies is created in `create_fresh_orca_install`.

rizbihassan: tried the above method. Check this 
https://github.com/nasa/cumulus-orca/pull/387

AndrewDorn (approved):

PR #413, opened 2023-05-10
Feature/orca 635 Delete old data from s3 and ORCA catalog after integration tests

rizbihassan:
## Summary of Changes

Addresses [ORCA-635: Delete old data from s3 and ORCA catalog after integration tests](https://bugs.earthdata.nasa.gov/browse/ORCA-635)

## Changes

* Created integration_test/cleanup/test-cleanup.py
* Updated readme.md
* Added delete action to cross account policy

## Type of change

- [x ] This change requires a documentation update

## How Has This Been Tested?

Ran the script locally and verified in aws.

## Checklist:

    - [ x] Testing documentation
- [x ] My changes generate no new warnings
- [x ] I have added tests that prove my fix is effective or that my feature works
    - [ x] Integration tests
    - [x ] Snyk
    - [ x] git-secrets

### DIFF
--- a/integration_test/cleanup/test-cleanup.py
+++ b/integration_test/cleanup/test-cleanup.py
@@ -1,0 +1,105 @@
+import boto3
+import os
+from orca_shared.database import shared_db
+from sqlalchemy import text
+from sqlalchemy.future import Engine
+from aws_lambda_powertools import Logger
+from botocore.exceptions import ClientError
+
+
+LOGGER = Logger()
+
+# os.env keys
+recovery_bucket_name = os.environ["orca_RECOVERY_BUCKET_NAME"]
+source_bucket_name = os.environ["SOURCE_BUCKET_NAME"]
+
+# grab the objects to be deleted
+s3 = boto3.client('s3')
+result = s3.list_objects_v2(
+    Bucket=source_bucket_name,
+    Prefix='test'
+)
+filenames = []
+for item in result['Contents']:
+    files = item['Key']
+    filenames.append(files)
+LOGGER.info(filenames)
+
+# Delete objects from source and recovery buckets
+for object in filenames:
+    response_source = s3.delete_object(
+        Bucket=source_bucket_name,
+        Key=object,
+    )
+    response_destination = s3.delete_object(
+        Bucket=recovery_bucket_name,
+        Key=object,
+    )
+
+# verify objects are deleted from buckets
+for object in filenames:
+    try:
+        head_object_output = s3.head_object(
+        Bucket=source_bucket_name,Key=object
+        )
+    except ClientError as error:
+        if error.response['Error']['Code'] == '404':
+            LOGGER.info(f"{object} not found in {source_bucket_name}. Test passed")
+for object in filenames:
+    try:
+        head_object_output = s3.head_object(
+        Bucket=recovery_bucket_name,Key=object
+        )
+    except ClientError as error:
+        if error.response['Error']['Code'] == '404':
+            LOGGER.info(f"{object} not found in {recovery_bucket_name}. Test passed")
+
+#delete from catalog
+try:
+    db_connect_info_secret_arn = os.environ["DB_CONNECT_INFO_SECRET_ARN"]
+except KeyError:
+    LOGGER.error("DB_CONNECT_INFO_SECRET_ARN environment value not found.")
+    raise
+db_connect_info = shared_db.get_configuration(db_connect_info_secret_arn)
+engine = shared_db.get_user_connection(db_connect_info)
+print(engine)
+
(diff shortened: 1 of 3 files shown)


integration_test/dr-buckets.tf.template
```diff
@@ -184,7 +184,8 @@ data "aws_iam_policy_document" "orca-primary-policy-json" {
                 "s3:GetBucket*",
                 "s3:ListBucket",
                 "s3:PutObjectAcl",
-                "s3:PutBucketNotification"
+                "s3:PutBucketNotification",
+                "s3:DeleteObject"
```
rizbihassan, line 188: This is needed to delete objects from the DR account bucket

integration_test/cleanup/test-cleanup.py
```diff
@@ -0,0 +1,77 @@
...
+LOGGER.info(filenames)
+
+# Delete objects from source and recovery buckets
+for object in filenames:
+    response_source = s3.delete_object(
+        Bucket=source_bucket_name,
+        Key=object,
+    )
+    response_destination = s3.delete_object(
+        Bucket=recovery_bucket_name,
```
AndrewDorn, line 35: This does assume that the files will have the same keys in each bucket. Fine for now, but it might be wise to future-proof by searching for all files present in the recovery bucket and appending them to a different list.

integration_test/cleanup/test-cleanup.py
```diff
@@ -0,0 +1,77 @@
...
+        Key=object,
+    )
+
+# verify objects are deleted from buckets
+for object in filenames:
+    try:
+        head_object_output = s3.head_object(
+        Bucket=source_bucket_name,Key=object
+        )
```
AndrewDorn, line 44: I like the logging here.

AndrewDorn: There doesn't seem to be a failure state... As-in, if the expected ClientError is not raised, the user is not notified and no error is raised.

integration_test/cleanup/test-cleanup.py
```diff
@@ -0,0 +1,77 @@
...
+        head_object_output = s3.head_object(
+        Bucket=recovery_bucket_name,Key=object
+        )
+    except ClientError as error:
+        if error.response['Error']['Code'] == '404':
+            LOGGER.info(f"{object} deleted from {recovery_bucket_name}. Test passed!")
+
+#delete from ORCA catalog
+db_connect_info = shared_db.get_configuration(db_connect_info_secret_arn)
+db_connect_info["host"]= "localhost"  # this is needed when running locally
```
AndrewDorn, line 59: Random missing space. `host"]= "localhost`

integration_test/cleanup/test-cleanup.py
```diff
@@ -0,0 +1,77 @@
...
+            LOGGER.info(f"{object} deleted from {source_bucket_name}. Test passed!")
+for object in filenames:
+    try:
+        head_object_output = s3.head_object(
+        Bucket=recovery_bucket_name,Key=object
+        )
+    except ClientError as error:
+        if error.response['Error']['Code'] == '404':
+            LOGGER.info(f"{object} deleted from {recovery_bucket_name}. Test passed!")
+
```
AndrewDorn, line 56: Good strong division here between the two sections.

AndrewDorn (changes requested):
Few minor optimizations. Overall looks great.

integration_test/cleanup/test-cleanup.py
rizbihassan: added failure state

integration_test/cleanup/test-cleanup.py
rizbihassan: so you are saying separately grab the files under "test" dir in a new list in recovery bucket and then do the deletion?

rizbihassan: updated with creating a new list for recovery bucket

integration_test/cleanup/test-cleanup.py
AndrewDorn: Close. I'd recommend removing the `if head_recovery_object_output["ResponseMetadata"]["HTTPStatusCode"] == 200:` check. It adds a side-path that we don't expect, and don't have any code to handle.
```
for object in recovery_filenames:
    try:
        head_recovery_object_output = s3.head_object(
        Bucket=recovery_bucket_name,Key=object
        )
        raise Exception(f"{object} still exists in {recovery_bucket_name}")    
    except ClientError as error:
        if error.response['Error']['Code'] == '404':
            LOGGER.info(f"{object} deleted from {recovery_bucket_name}. Test passed!")
        else:
            raise error
```

AndrewDorn (approved):

PR #457, opened 2023-10-12
Feature/orca 737 Update moto library to 4.2.2 and make necessary fixes in unit tests

rizbihassan:
## Summary of Changes

Addresses [ORCA-737: Update moto library to 4.2.2 and make necessary fixes in unit tests](https://bugs.earthdata.nasa.gov/browse/ORCA-737)

## Changes

* Updated moto library version to 4.2.2
* Updated relevant unit tests that have been affected by moto updates.

## Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Ran unit tests locally. Deployed in AWS.

## Checklist:

    - [x ] CHANGELOG.md
- [x ] I have added tests that prove my fix is effective or that my feature works
    - [ x] Unit tests

    - [x ] Snyk
    - [x ] git-secrets

### DIFF
--- a/shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
+++ b/shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
@@ -30,18 +30,12 @@ class TestSharedRecoveryLibraries(unittest.TestCase):
         """
         self.mock_sqs.start()
         self.test_sqs = boto3.resource("sqs", region_name="us-west-2")
-        self.queue = self.test_sqs.create_queue(QueueName="unit-test-queue")
-        self.db_queue_url = self.queue.url
-        self.request_methods = [
-            shared_recovery.RequestMethod.NEW_JOB,
-            shared_recovery.RequestMethod.UPDATE_FILE,
-        ]
-        self.statuses = [
-            shared_recovery.OrcaStatus.PENDING,
-            shared_recovery.OrcaStatus.STAGED,
-            shared_recovery.OrcaStatus.SUCCESS,
-            shared_recovery.OrcaStatus.FAILED,
-        ]
+        self.standard_queue = self.test_sqs.create_queue(QueueName="standard-queue")
+        self.standard_queue_url = self.standard_queue.url
+        self.fifo_queue = self.test_sqs.create_queue(
+            QueueName="fifo-queue.fifo", Attributes={"FifoQueue": "true"}
+        )
+        self.fifo_queue_url = self.fifo_queue.url
         self.job_id = "1234"
         self.granule_id = "6c8d0c8b-4f9a-4d87-ab7c-480b185a0250"
         self.MessageGroupId = "request_files"
@@ -64,22 +58,18 @@ class TestSharedRecoveryLibraries(unittest.TestCase):
         function returns the same expected message.
         """
         new_data = {"name": "test"}
-        for request_method in self.request_methods:
-            # Run subtests
-            with self.subTest(request_method=request_method):
-                # Send values to the function
-                shared_recovery.post_entry_to_fifo_queue(
-                    new_data, request_method, self.db_queue_url
-                )
+        # Run subtests
+        # Send values to the function
+        shared_recovery.post_entry_to_fifo_queue(
+            new_data, shared_recovery.RequestMethod.NEW_JOB, self.fifo_queue_url
+        )
 
-                # grabbing queue contents after the message is sent
-                queue_contents = self.queue.receive_messages(
-                    MessageAttributeNames=["All"]
-                )
-                queue_output_body = json.loads(queue_contents[0].body)
+        # grabbing queue contents after the message is sent
+        queue_contents = self.fifo_queue.receive_messages(MessageAttributeNames=["All"])
+        queue_output_body = json.loads(queue_contents[0].body)
 
-                # Testing SQS body
-                self.assertEqual(queue_output_body, new_data)
+        # Testing SQS body
+        self.assertEqual(queue_output_body, new_data)
 
     @patch.dict(
         os.environ,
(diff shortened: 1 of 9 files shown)


shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
```diff
@@ -64,22 +58,18 @@ def test_post_entry_to_fifo_queue_no_errors(self):
...
         new_data = {"name": "test"}
-        for request_method in self.request_methods:
-            # Run subtests
-            with self.subTest(request_method=request_method):
-                # Send values to the function
-                shared_recovery.post_entry_to_fifo_queue(
-                    new_data, request_method, self.db_queue_url
-                )
+        # Run subtests
+        # Send values to the function
```
bhazuka, line 62: So originally we tested both 
```
self.request_methods = [
            shared_recovery.RequestMethod.NEW_JOB,
            shared_recovery.RequestMethod.UPDATE_FILE,
        ]
```

Any reason why we are only testing `NEW_JOB` now?

shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
```diff
@@ -157,88 +149,110 @@ def test_create_status_for_job_no_errors(self):
         {"AWS_REGION": "us-west-2"},
         clear=True,
     )
-    def test_update_status_for_file_no_errors(self):
+    def test_update_status_for_file_no_errors_status_success(self):
```
bhazuka, line 152: Not really sure why this was broken into 4 tests seems like a lot of overhead to manage when the input and output is always expected to be the same with the exception of and ERROR status. At the most I could see 2 tests here but the 1 test we had before to cycle through the values seems more efficient and easy to maintain. Is there a good reason for the breakout?

bhazuka (changes requested):
A couple of items to talk through and understand.

shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
rizbihassan: Unfortunately FIFO queue does not work with subtests. If we use the same input for different tests using  `NEW_JOB` and `UPDATE_FILE `, it throws an error. 

```
 IndexError: list index out of range
```

My assumption is since we are sending the same input to FIFO queue, it only accepts the first message and ignores the one ones.

shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
rizbihassan: Unfortunately FIFO queue does not work with subtests. If we use the same input for different tests using  the four statuses, it throws an error. 

```
 IndexError: list index out of range
```

My assumption is since we are sending the same input to FIFO queue, it only accepts the first message and ignores the one ones.

zroehrich (approved):

shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
bhazuka: So I did a little digging and figured out what was going on. So the updates to moto actually put it closer in reality to what an SQS queue does. So in order to better test, here are the modifications I made to the original test file. Note that I am only showing the additions and a few reference lines, not the whole code.

```python
def setUp(self):
    """
    Perform initial setup for the tests.
    """
    ...
    self.fifo_queue = self.test_sqs.create_queue(
        QueueName="unit-test-fifo-queue.fifo", Attributes={"FifoQueue": "true"}
    )
    self.fifo_queue_url = self.fifo_queue.url
    ...

def test_post_entry_to_fifo_queue_no_errors(self):
        """
        *Happy Path*
        Test that sending a message to SQS queue using post_entry_to_fifo_queue()
        function returns the same expected message.
        """
        new_data = {"name": "test"}
        for request_method in self.request_methods:
            with self.subTest(request_method=request_method):
                # Send values to the function
                shared_recovery.post_entry_to_fifo_queue(
                    new_data, request_method, self.fifo_queue_url
                )
                queue_contents = self.fifo_queue.receive_messages(
                    MessageAttributeNames=["All"]
                )
                
                # ADDING test to verify message attribute
                self.assertEqual(
                    request_method.value,
                    queue_contents[0].message_attributes["RequestMethod"][
                        "StringValue"
                    ],
                )
                queue_output_body = json.loads(queue_contents[0].body)

                # Testing SQS body
                self.assertEqual(queue_output_body, new_data)
                
                # DELETE the message from the queue to show we have read it
                self.fifo_queue.delete_messages(
                    Entries=[
                        {
                            "ReceiptHandle": queue_contents[0].receipt_handle,
                            "Id": queue_contents[0].message_id,
                        }
                    ]
                )

```

shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
bhazuka: See comments from previous comment. We should be able to use subtests with fifo queue, we just need to add a delete message before the loop goes around again.

```python
self.fifo_queue.delete_messages(
                    Entries=[
                        {
                            "ReceiptHandle": queue_contents[0].receipt_handle,
                            "Id": queue_contents[0].message_id,
                        }
                    ]
                )
```

shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
rizbihassan: This works! I forgot that the message needs to be read first in FIFO before going to the next message. Good to know moto is smart to handle that!

shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
rizbihassan: Yup this works!

bhazuka (changes requested):
Add 1 additional test in based on comment.

shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
```diff
@@ -46,30 +56,39 @@ def tearDown(self):
...
+                queue_contents = self.fifo_queue.receive_messages(
+                    MessageAttributeNames=["All"]
+                )
+                queue_output_body = json.loads(queue_contents[0].body)
 
-        # Testing SQS body
-        self.assertEqual(queue_output_body, new_data)
+                # Testing SQS body
+                self.assertEqual(queue_output_body, new_data)
+
```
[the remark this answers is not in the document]
rizbihassan: Added

bhazuka (approved):

PR #460, opened 2023-10-25
feature/ORCA-704 Remove region from automated integration testing

zroehrich:
## Summary of Changes

Addresses [ORCA-704: Remove region from automated integration testing](https://bugs.earthdata.nasa.gov/browse/ORCA-704)

## Changes

* Updated buckets.tf.template with an AWS provider block to deploy to the us-west-2 region due to unsuccessful deployments in the other regions.

## Type of change

## How Has This Been Tested?
Deployed the new template to AWS for validation.

## Checklist:

    - [x] git-secrets
    - [x] Snyk


integration_test/buckets.tf.template
```diff
@@ -1,5 +1,9 @@
 data "aws_caller_identity" "current_account" {}
 
+provider "aws" {
+  region = "us-west-2"
+}
+
```
rizbihassan, line 6: If we are defaulting to us-west-2, we should also do it in dr-buckets.tf.template.

Also I think default_region can be removed from the bamboo spec files that deploy the buckets.

rizbihassan (changes requested):

integration_test/buckets.tf.template
rizbihassan: Verify in bamboo-specs/specs/deploy-cumulus-orca-buckets.yaml

zroehrich: Added to the dr-buckets.tf.template. Not sure if we want to remove the default_region variable from the bamboo spec files since the scripts used in that reference that variable for ARNs

rizbihassan: If you have access to bamboo CI, it can be tested without the region to see if it works.

bhazuka (approved):

bamboo-specs/specs/deploy-cumulus-orca-buckets.yaml
```diff
@@ -98,8 +98,8 @@ Run Ingest Integration tests:
 
 variables:
   DEPLOYMENT: sandbox
-  AWS_DEFAULT_REGION: us-west-2
-  CUMULUS_AWS_DEFAULT_REGION: us-west-2
+  #AWS_DEFAULT_REGION: us-west-2
+  #CUMULUS_AWS_DEFAULT_REGION: us-west-2
```
rizbihassan, line 102: I tested and this will throw an error if commented out. Recommend removing the # for now.

rizbihassan (approved):
Looks good. Users don't need to update/change the AWS_DEFAULT_REGION because it's also hardcoded in `bamboo-specs/specs/deploy-cumulus-orca-buckets.yaml` variable section. We can keep the variable there for now but don't need to use it.

PR #471, opened 2023-11-30
ORCA-366 - Improve/Add additional unit tests needed in shared_recovery.py

zroehrich:
## Summary of Changes

Addresses [ORCA-366: Improve/Add additional unit tests needed in shared_recovery.py.](https://bugs.earthdata.nasa.gov/browse/ORCA-366)

## Changes

* Added unit test to verify request_method has the correct value associated

## Type of change

## How Has This Been Tested?
Ran using Bamboo and locally.

## Checklist:

    - [x] git-secrets
    - [x] Snyk

### DIFF
--- a/shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
+++ b/shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
@@ -94,6 +94,7 @@ class TestSharedRecoveryLibraries(unittest.TestCase):
                         }
                     ]
                 )
+                print(request_method.value)
 
     @patch.dict(
         os.environ,
--- a/shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
+++ b/shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
@@ -94,7 +94,12 @@ class TestSharedRecoveryLibraries(unittest.TestCase):
                         }
                     ]
                 )
-                print(request_method.value)
+                # Verifies that request_method is set properly (new_job or update_file)
+                self.assertTrue(
+                    request_method.value == "new_job"
+                    or request_method.value == "update_file",
+                    f"Incorrect Request Method: {request_method.value}",
+                )
 
     @patch.dict(
         os.environ,


shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
```diff
@@ -94,6 +94,12 @@ def test_post_entry_to_fifo_queue_no_errors(self):
                         }
                     ]
                 )
+                # Verifies that request_method is set properly (new_job or update_file)
+                self.assertTrue(
+                    request_method.value == "new_job"
```
rizbihassan, line 99: Shouldn't this be specific to either `new_job` or `update_file`?

zroehrich: So this will check if request_method is set as `new_job` or `update_file`, as long as request_method has one of those set the test will pass, but if request_method is set to something else like `delete_file` the test will fail and output a message with what the incorrect value is.

rizbihassan: I think we should also have a test that fails in case of `request_method` equals to something else other than `new_job` or `update_file`

rizbihassan (changes requested):

shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
zroehrich: This test will fail if `request_method` is set to something else other than `new_job` or `update_file`. Tested by adding new request methods and `request_method` was set to `delete_file` and the test failed:
`AssertionError: False is not true : Incorrect Request Method: delete_file`

Bamboo output: https://ci.earthdata.nasa.gov/browse/ORCA-OI405-JOB1-5/log#:~:text=AssertionError%3A%20False%20is%20not%20true%20%3A%20Incorrect%20Request%20Method%3A%20delete_file

So if `request_method` is set to `new_job` or `update_file` the test will pass, and if it's equal to anything else the test will fail.

rizbihassan: Understood but lets add that in a new test which uses self.assertRaises or something similar to verify that it is indeed failing.

Something like: 
 test_request_method_raises_error(self)
...........

shared_libraries/orca_shared/recovery/shared_recovery.py
```diff
@@ -271,5 +271,5 @@ def post_entry_to_standard_queue(
 
 
 def get_request_method(request_method: RequestMethod):
-    if request_method != "new_job":
+    if request_method != "new_job" or request_method != "update_file":
```
rizbihassan, line 274: We should not change the original python file (shared_recovery.py in this case) when doing unit tests. All changes should be done in the test file.

shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
```diff
@@ -281,3 +287,9 @@ def test_update_status_for_file_error_message_empty_raises_error_message(self):
                     uuid.uuid4().__str__(),
                 )
             self.assertEqual("Error message is required.", str(cm.exception))
+
+    def test_request_method_raises_error(self):
+        for request_method in self.request_methods:
+            with self.assertRaises(ValueError) as cm:
+                shared_recovery.get_request_method(request_method.value)
+            self.assertEqual(str(cm.exception), "new_job")
```
rizbihassan, line 295: Remove this test.

shared_libraries/orca_shared/recovery/test/unit_tests/test_shared_recovery.py
```diff
@@ -94,6 +94,12 @@ def test_post_entry_to_fifo_queue_no_errors(self):
                         }
                     ]
                 )
+                # Verifies that request_method is set properly to new_job or update_file
+                self.assertTrue(
+                    request_method.value == "new_job"
```
rizbihassan, line 99: Looks like we already have test to check request_method 

```
                self.assertEqual(
                    request_method.value,
                    queue_contents[0].message_attributes["RequestMethod"][
                        "StringValue"
                    ],
                )

```


So you can keep this code and I think we are good.

shared_libraries/orca_shared/recovery/shared_recovery.py
zroehrich: Undid change

PR #478, opened 2024-01-04
ORCA-787: Update ORCA API pathing by removing the extra `orca` to avoid confusion.

zroehrich:
## Summary of Changes

Addresses [ORCA-787: Update ORCA API pathing by removing the extra `orca` to avoid confusion.](https://bugs.earthdata.nasa.gov/browse/ORCA-787)

## Changes

* Updated ORCA API pathing by removing the extra `orca` to avoid confusion.

## Type of change

## How Has This Been Tested?
Tested API path in AWS console

## Checklist:

    - [x] git-secrets
    - [x] Snyk

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and includes an additional section for migration notes.
 - *ORCA-784* Changed documentation to replace restore with copy based on task's naming as well as changed file name from `website/docs/operator/restore-to-orca.mdx` to `website/docs/operator/reingest-to-orca.mdx`.
 - *ORCA-724* Updated ORCA recovery documentation to include recovery workflow process and relevant inputs and outputs in `website/docs/operator/data-recovery.md`.
 - *ORCA-789* Updated `extract_filepaths_for_granule` to more flexibly match file-regex values to keys.
+- *ORCA-787* Updated `modules/api-gateway/main.tf` to remove extra `orca` in path to avoid confusion.
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ and includes an additional section for migration notes.
 - *ORCA-784* Changed documentation to replace restore with copy based on task's naming as well as changed file name from `website/docs/operator/restore-to-orca.mdx` to `website/docs/operator/reingest-to-orca.mdx`.
 - *ORCA-724* Updated ORCA recovery documentation to include recovery workflow process and relevant inputs and outputs in `website/docs/operator/data-recovery.md`.
 - *ORCA-789* Updated `extract_filepaths_for_granule` to more flexibly match file-regex values to keys.
-- *ORCA-787* Updated `modules/api-gateway/main.tf` to remove extra `orca` in path to avoid confusion.
+- *ORCA-787* Modified `modules/api-gateway/main.tf` api gateway stage name to remove the extra orca from the data management URL path
 
 ### Deprecated
 
--- a/modules/api-gateway/variables.tf
+++ b/modules/api-gateway/variables.tf
@@ -54,7 +54,7 @@ variable "orca_catalog_reporting_invoke_arn" {
 variable "api_gateway_stage_name" {
   type        = string
   description = "stage name for the ORCA cumulus reconciliation api gateway"
-  default     = "orca"
+  default     = "api"
 }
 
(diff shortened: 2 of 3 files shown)


rizbihassan (changes requested):
Have you tested deploying this change in our orca deployment in aws? I got this error while deploying your branch

`
╷
│ Error: updating API Gateway Resource (rwqful): NotFoundException: Invalid Resource identifier specified
│
│   with module.orca.module.orca.module.orca_api_gateway.aws_api_gateway_resource.orca_datamanagement_api_resource,
│   on ../../cumulus-orca/modules/api-gateway/main.tf line 237, in resource "aws_api_gateway_resource" "orca_datamanagement_api_resource":
│  237: resource "aws_api_gateway_resource" "orca_datamanagement_api_resource" {
│
`

This is the path that has the extra `orca` which needs to be removed and tested.
Screenshot 2024-01-05 at 8 24 00 AM


Once that is deployed, you need to trigger the APIs with the correct input in aws to see if it gives any error.

zroehrich:
I ran into that same error but it goes away if you run an apply a second time. Not sure why it does that

rizbihassan:
> I ran into that same error but it goes away if you run an apply a second time. Not sure why it does that

Yes got a successful deployment on 2nd try. However, the URL still shows the duplicate orca in path. 

After further digging, it looks like the extra `orca` is actually the stage name in the deployment. So we should change the stage name from orca to something else like `api`

zroehrich:
So do we want to add back in the `orca` path right before `datamanagement` and change the stage name or keep it as is and change the stage name to `api`?

rizbihassan:
> So do we want to add back in the `orca` path right before `datamanagement` and change the stage name or keep it as is and change the stage name to `api`?

Ya I think add back the orca path. 
You  can change the stage name here https://github.com/nasa/cumulus-orca/blob/62c75c05a77da676cf247d3e2b49bc2a30077a15/modules/api-gateway/variables.tf#L57 and see if this fixes the path.

modules/api-gateway/variables.tf
```diff
@@ -54,7 +54,7 @@ variable "orca_catalog_reporting_invoke_arn" {
 variable "api_gateway_stage_name" {
   type        = string
   description = "stage name for the ORCA cumulus reconciliation api gateway"
-  default     = "orca"
+  default     = "api"
```
rizbihassan, line 57: I now get this error while deploying 

```
│ Error: deleting API Gateway Deployment (dbtt7b): BadRequestException: Active stages pointing to this deployment must be moved or deleted

```

rizbihassan (changes requested):

modules/api-gateway/variables.tf
rizbihassan: I still get the same error. Does this change update the API pathing I posted above in the screenshot?

zroehrich: Yes it will change it from `https://<api-id>.execute-api.us-west-2.amazonaws.com/orca/orca/datamanagement/reconciliation/internal/jobs` to `https://<api-id>.execute-api.us-west-2.amazonaws.com/api/orca/datamanagement/reconciliation/internal/jobs`

rizbihassan: Not sure why I am getting the stage error. Might be related to https://bugs.earthdata.nasa.gov/browse/ORCA-769

If you tested and verified that this works, then I think we are good.

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and includes an additional section for migration notes.
 - *ORCA-784* Changed documentation to replace restore with copy based on task's naming as well as changed file name from `website/docs/operator/restore-to-orca.mdx` to `website/docs/operator/reingest-to-orca.mdx`.
 - *ORCA-724* Updated ORCA recovery documentation to include recovery workflow process and relevant inputs and outputs in `website/docs/operator/data-recovery.md`.
 - *ORCA-789* Updated `extract_filepaths_for_granule` to more flexibly match file-regex values to keys.
+- *ORCA-787* Updated `modules/api-gateway/main.tf` to remove extra `orca` in path to avoid confusion.
```
rizbihassan, line 29: Maybe update to "Modified api gateway stage name to remove the extra `orca` from the data management URL path"

rizbihassan (approved):

PR #492, opened 2024-02-01
ORCA-826: Modify Delete Lambda Scripts to Delete on Tags

zroehrich:
## Summary of Changes

Addresses [ORCA-826: Modify Delete Lambda Scripts to Delete on Tags](https://bugs.earthdata.nasa.gov/browse/ORCA-826)

## Changes

* Modified `bin/delete_lambda.py` to delete ORCA lambdas based on their tags

## Type of change

## How Has This Been Tested?
Ran script and deleted only ORCA lambdas with the matching tag and not any other lambdas with the same prefix.

## Checklist:

    - [x] git-secrets
    - [x] Snyk

### DIFF
--- a/bin/delete_lambda.py
+++ b/bin/delete_lambda.py
@@ -1,7 +1,7 @@
 import subprocess
 import json
 
-# Gets user input for the prefix of ORCA lambdas to delete
+# Gets user input for the prefix of lambdas to delete
 prefix = input("Enter Prefix: ")
 
 # Gets ORCA lambda functions with given prefix
@@ -19,11 +19,13 @@ for sub in convert:
     try:
         # Finds lambdas with the tag application set to ORCA and deletes the ORCA Lambdas
         for tag in tag_convert:
-            if tag_convert[tag]['application'] == 'ORCA':
-                print("Deleting " + sub['FunctionName'])
-                del_lambda = sub['FunctionName']
-                delete_function = f"aws lambda delete-function --function-name {del_lambda}"
-                subprocess.run(delete_function, shell=True, capture_output=True)
-    # If the lambda with the provided prefix does not match the tag for deletion it is skipped and is not deleted
-    except KeyError:
-        pass
+            for attr in tag_convert[tag]:
+                if attr == 'application':
+                    if tag_convert[tag]['application'] == 'ORCA':
+                        print("Deleting " + sub['FunctionName'])
+                        del_lambda = sub['FunctionName']
+                        delete_function = f"aws lambda delete-function --function-name {del_lambda}"
+                        subprocess.run(delete_function, shell=True, capture_output=True)
+    # Throws an error if deletion is interrupted
+    except KeyError as ke:
+        raise ke
--- a/bin/delete_lambda.py
+++ b/bin/delete_lambda.py
@@ -10,9 +10,20 @@ completed_process = subprocess.run(get_functions, shell=True, capture_output=Tru
 output = completed_process.stdout
 convert = json.loads(output.decode("utf-8").replace("'","'"))
 
-# Deletes ORCA lambda functions with the given prefix
 for sub in convert:
-    print("Deleting " + sub['FunctionName'])
-    lambda_output = sub['FunctionName']
-    delete_function = f"aws lambda delete-function --function-name {lambda_output}"
-    subprocess.run(delete_function, shell=True, capture_output=True)
+    lambda_output = sub['FunctionArn']
+    get_tags = f"aws lambda list-tags --resource {lambda_output}"
+    tags_process = subprocess.run(get_tags, shell=True, capture_output=True)
+    tag_output = tags_process.stdout
+    tag_convert = json.loads(tag_output.decode("utf-8").replace("'","'"))
+    try:
+        # Finds lambdas with the tag application set to ORCA and deletes the ORCA Lambdas
+        for tag in tag_convert:
+            if tag_convert[tag]['application'] == 'ORCA':
+                print("Deleting " + sub['FunctionName'])
+                del_lambda = sub['FunctionName']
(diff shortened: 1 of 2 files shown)


bin/delete_lambda.py
```diff
@@ -10,9 +10,20 @@
...
+        # Finds lambdas with the tag application set to ORCA and deletes the ORCA Lambdas
+        for tag in tag_convert:
+            if tag_convert[tag]['application'] == 'ORCA':
+                print("Deleting " + sub['FunctionName'])
+                del_lambda = sub['FunctionName']
+                delete_function = f"aws lambda delete-function --function-name {del_lambda}"
+                subprocess.run(delete_function, shell=True, capture_output=True)
+    # If the lambda with the provided prefix does not match the tag for deletion it is skipped and is not deleted
+    except KeyError:
+        pass
```
rizbihassan, line 29: If for some reason the delete API does not work, it will ignore the error. Recommend changing `pass` to raising an exception.

rizbihassan (changes requested):
For some reason, running the script does not delete the lambdas for me.

```
(venv) IGSKMNCNLTM0057:cumulus-orca rhassan$ python3 bin/delete_lambda.py 
Enter Prefix: rhassanorca1
(venv) IGSKMNCNLTM0057:cumulus-orca rhassan$ 
```

bin/delete_lambda.py
zroehrich: Do you have the default profile set up for the cumulus account, I was able to find the lambdas matching that prefix and tag (took out the deletion when I ran the script just to list them and not delete). Alternatively I could add an input for a profile to use incase users have the default profile set for another account and can specify a cumulus profile
```
zroehrich@IGSKMN011LM0069 ~/Desktop/s3/test1 -  $ python3 delete.py
Enter Prefix: rhassanorca1
Deleting rhassanorca1_db_deploy
Deleting rhassanorca1_extract_filepaths_for_granule
Deleting rhassanorca1_post_to_database
Deleting rhassanorca1_post_copy_request_to_queue
Deleting rhassanorca1_internal_reconcile_report_orphan
Deleting rhassanorca1_delete_old_reconcile_jobs
Deleting rhassanorca1_request_from_archive
Deleting rhassanorca1_internal_reconcile_report_phantom
Deleting rhassanorca1_internal_reconcile_report_job
Deleting rhassanorca1_copy_from_archive
Deleting rhassanorca1_orca_catalog_reporting
Deleting rhassanorca1_copy_to_orca
Deleting rhassanorca1_request_status_for_job
Deleting rhassanorca1_internal_reconcile_report_mismatch
Deleting rhassanorca1_get_current_archive_list
Deleting rhassanorca1_perform_orca_reconcile
Deleting rhassanorca1_post_to_queue_and_trigger_step_function
Deleting rhassanorca1_request_status_for_granule
Deleting rhassanorca1_post_to_catalog
zroehrich@IGSKMN011LM0069 ~/Desktop/s3/test1 -  $ 
```

rizbihassan: My bad! I was connected to the DR account instead of the cumulus account which is why it could not find anything to delete with that prefix.

Looks good now.

zroehrich: Okay I added in an input for an AWS profile to use so if users want to use a profile outside of default they can do that.
```
zroehrich@IGSKMN011LM0069 ~/Desktop/s3/test1 -  $ python3 delete.py
Enter Prefix: zr4
Enter AWS Profile: cumulus-sandbox
Deleting zr4_internal_reconcile_report_orphan
Deleting zr4_post_to_database
Deleting zr4_post_to_catalog
Deleting zr4_orca_catalog_reporting
Deleting zr4_request_status_for_job
Deleting zr4_perform_orca_reconcile
Deleting zr4_delete_old_reconcile_jobs
Deleting zr4_post_to_queue_and_trigger_step_function
Deleting zr4_get_current_archive_list
Deleting zr4_extract_filepaths_for_granule
Deleting zr4_internal_reconcile_report_job
Deleting zr4_copy_to_orca
Deleting zr4_db_deploy
Deleting zr4_post_copy_request_to_queue
Deleting zr4_request_from_archive
Deleting zr4_internal_reconcile_report_phantom
Deleting zr4_copy_from_archive
Deleting zr4_request_status_for_granule
Deleting zr4_internal_reconcile_report_mismatch
zroehrich@IGSKMN011LM0069 ~/Desktop/s3/test1 -  $ 
```

bin/delete_lambda.py
```diff
@@ -1,18 +1,21 @@
 import subprocess
 import json
 
-# Gets user input for the prefix of lambdas to delete
+# Gets user input for the prefix of ORCA lambdas to delete
 prefix = input("Enter Prefix: ")
 
+# Gets AWS Profile to use for ORCA lambda deletion
+profile = input("Enter AWS Profile: ")
```
rizbihassan, line 8: I am confused why we need a profile here. I think this can be removed.

zroehrich: Removed

bin/delete_lambda.py
rizbihassan, line 29: I think if we remove `pass` and raise an exception, it will throw an error in case the deletion operation is interrupted or if tag is not present which is what we want. Otherwise, it will fail silently.
```
except KeyError as ke:
        raise ke
```

rizbihassan (changes requested):

bin/delete_lambda.py
zroehrich: Added that in, modified the script a bit to do some more comparisons because the way it was functioning is if it found a lambda without that `application` tag it would error out before deleting the ORCA lambdas

rizbihassan: Looks good

rizbihassan (approved):

PR #514, opened 2024-05-24
ORCA-868: Create Temporary EC2 Instance for DB Comparison

zroehrich:
## Summary of Changes

Added temporary EC2 instance and instructions for DB comparison after the upgrade/migration to Aurora Serverless V2

Addresses [ORCA-868: Create Temporary EC2 Instance for DB Comparison](https://bugs.earthdata.nasa.gov/browse/ORCA-868)

## Changes

* Added EC2 Terraform
* Added psql install script
* Added variables file for compare script
* Added Instructions for deployment and use of EC2

## Type of change

## How Has This Been Tested?

Deployed and tested within AWS

## Checklist:

    - [x ] CHANGELOG.md
    - [x ] User documentation
    - [ x] Snyk
    - [ x] git-secrets


rizbihassan (commented):
Looks good. 
I am wondering if we can add any initial troubleshooting steps for users in case the count in v1 and v2 tables do not match.

zroehrich:
> Looks good. I am wondering if we can add any initial troubleshooting steps for users in case the count in v1 and v2 tables do not match.

I was thinking about that but will be a very case by case thing on how they'd want to handle it, depending on how many records are missing they may want to completely redo the migration with a new snapshot, a pg_dump, re-reunning workflows, etc. Ideally they would want to consult with like a DBA on best steps to go with since if there are differences every user's difference's will be the same, where some might be missing just a couple of records someone else may be missing hundreds or thousands.

rizbihassan:
> > Looks good. I am wondering if we can add any initial troubleshooting steps for users in case the count in v1 and v2 tables do not match.
> 
> I was thinking about that but will be a very case by case thing on how they'd want to handle it, depending on how many records are missing they may want to completely redo the migration with a new snapshot, a pg_dump, re-reunning workflows, etc. Ideally they would want to consult with like a DBA on best steps to go with since if there are differences every user's difference's will be the same, where some might be missing just a couple of records someone else may be missing hundreds or thousands.

Makes sense.

rizbihassan (approved):

